PR #18, opened 2020-10-02
Sarif converter 

Spoor2709:
We should discuss the partial footprint as well as the related locations in more detail.


src/analysis.ts
```diff
@@ -273,12 +275,26 @@ export async function analyzeGit(
...
+  }
+  else {
+    return {
+      baseURL,
+      sessionToken,
+      includeLint,
+      severity,
+      gitUri,
+      ...analysisData,
+      sarifResults: analysisData.analysisResults ? await new Sarif(analysisData.analysisResults).sarifConverter() : undefined,
```
shinin-light, line 296: Why creating a new instance of the class for each conversion? "await" and "new" both have their own computational weight which could be easily avoided by using static functions.

src/sarif_converter.ts
```diff
@@ -0,0 +1,180 @@
...
+  file: string;
+}
+interface ISarifSuggestions {
+  [suggestionIndex: number]: ISarifSuggestion;
+}
+class Sarif {
+  private analysisResults: IAnalysisResult;
+  private suggestions: ISarifSuggestions= {}
+
+  constructor(analysisResults: IAnalysisResult) {
```
shinin-light, line 19: Conceptually speaking, there is no meaning in the "Sarif" class being dependant/built around a single instance of the analysisResults.

src/sarif_converter.ts
```diff
@@ -0,0 +1,180 @@
...
+    this.analysisResults = analysisResults;
+    for (const [file] of Object.entries(this.analysisResults.files)) {
+      for (const [issueId, issue] of <[string, IFileSuggestion][]>Object.entries(this.analysisResults.files[file])) {
+        if (!this.suggestions || !Object.keys(this.suggestions).includes(issueId)) {
+          this.suggestions[issueId] = { ...issue[0], file: file.substring(1) };
```
shinin-light, line 24: Why do you need this step? Can't you just execute this code in the sarifConverter by passing the analysisResults as a parameter?

tests/git.analysis.spec.ts
```diff
@@ -40,5 +40,17 @@ describe('Functional test of analysis', () => {
...
+  it('analyze remote git with oid and return sarif', async () => {
+    const bundle = await analyzeGit(
+      baseURL,
+      sessionToken,
+      false,
+      1,
+      'git@github.com:DeepCodeAI/cli.git@320d98a6896f5376efe6cefefb6e70b46b97d566',
+      true,
+    );
+    expect(!!bundle.sarifResults).toEqual(true);
```
shinin-light, line 54: You would probably want to have some format specification here so that if we break something in the future the tests will tell us that the output format is wrong.

shinin-light (changes requested):
I didn't look at the format itself, since I'm not an expert of that space.

tests/git.analysis.spec.ts
Spoor2709: As discussed Type script already infers the output type and will not compile if we break something.

Arvi3d (changes requested):
Michael made a good about automated format testing of Sarif, please try to implement it.

shinin-light (approved):
Be sure to have some SARIF validator in the test text.

PR #26, opened 2020-11-04
Waiting status

Arvi3d:
- added waiting status to expected response statuses
- adjusted one test of analysis with missing files;
- fixed one test case with changed suggestion message

### DIFF
--- a/tests/api.spec.ts
+++ b/tests/api.spec.ts
@@ -194,20 +194,25 @@ describe('Requests to public API', () => {
   });
 
   it('request analysis with missing files', async () => {
-    const response = await getAnalysis({
-      baseURL,
-      sessionToken,
-      bundleId: fakeBundleIdFull,
-      includeLint: false,
-      severity: 1,
-    });
+    let response;
+    do {
+      response = await getAnalysis({
+        baseURL,
+        sessionToken,
+        bundleId: fakeBundleIdFull,
+        includeLint: false,
+        severity: 1,
+      });
+    } while (response.type === 'success');
+
     expect(response.type).toEqual('error');
-    if (response.type === 'success') return;
+    // if (response.type === 'success') return;
     expect(response.error).toEqual({
       apiName: 'getAnalysis',
       statusCode: 500,
       statusText: 'Getting analysis failed',
     });
+
   });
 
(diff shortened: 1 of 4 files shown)


tests/api.spec.ts
```diff
@@ -194,20 +194,24 @@ describe('Requests to public API', () => {
...
+    let response;
+    do {
+      response = await getAnalysis({
+        baseURL,
+        sessionToken,
+        bundleId: fakeBundleIdFull,
+        includeLint: false,
+        severity: 1,
+      });
+    } while (response.type === 'success');
```
shinin-light, line 206: I don't understand. Why should getAnalysis return 'success' for incomplete bundles? If this request loops once, it may be stuck forever until the bundle expires (1 day of test is not feasible).

tests/api.spec.ts
```diff
@@ -194,20 +194,24 @@ describe('Requests to public API', () => {
...
+
     expect(response.type).toEqual('error');
-    if (response.type === 'success') return;
     expect(response.error).toEqual({
       apiName: 'getAnalysis',
       statusCode: 500,
```
shinin-light, line 211: Shouldn't this be 404 instead of 500 for incomplete/expired bundle analysis in the new pubAPI queue branch?

tests/api.spec.ts
Arvi3d: Because in worst case, this request will have to wait to be added to the queue and we will not know if it's failed or not, so it will return success response.

Arvi3d: this is a test. Tests have timeouts themselves and will not pass in this case

tests/api.spec.ts
Arvi3d: hm, this is interesting question. let me check...

Arvi3d: fixed. Please check with together with this PR: https://github.com/DeepCodeAI/product/pull/643

tests/api.spec.ts
shinin-light: I don't know if the test timeout is a valid pattern for this scenario, but if you're sure it applies I'll resolve the conversation for now.

shinin-light (approved):

PR #30, opened 2020-11-27
new fields in suggestion:

Arvi3d:
- cwe
- title
- text

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -58,9 +58,6 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       shortDescription: {
         text: suggestion.title || suggestion.rule,
       },
-      fullDescription: {
-        text: suggestion.message,
-      },
       defaultConfiguration: {
         level: severity,
       },
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -60,7 +60,7 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       id: suggestionId,
       name: suggestion.rule,
       shortDescription: {
-        text: suggestion.title || suggestion.message,
+        text: suggestion.title || suggestion.rule,
       },
       fullDescription: {
         text: suggestion.message,
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -60,7 +60,7 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       id: suggestionId,
       name: suggestion.rule,
       shortDescription: {
-        text: suggestion.message,
+        text: suggestion.title || suggestion.message,
       },
       fullDescription: {
         text: suggestion.message,
@@ -69,10 +69,11 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
         level: severity,
       },
       help: {
+        markdown: suggestion.text,
         text: suggestion.message,
       },
       properties: {
-        tags: [suggestionId.split('%2F')[0], ...suggestion.tags, ...suggestion.categories],
+        tags: [suggestionId.split('%2F')[0], ...suggestion.cwe, ...suggestion.tags, ...suggestion.categories],
         precision: 'very-high',
       },
     };
@@ -86,7 +87,7 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       id: suggestionId,
       text: suggestion.message,
     };
-    ruleIndex++;
+    ruleIndex += 1;
   }
   return { tool: { driver: { ...output.driver, rules } }, suggestions };
 };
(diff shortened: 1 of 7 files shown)


deepcodeg:
@Arvi3d can't wait to see this in the IDEs!

orkamara (approved):

src/sarif_converter.ts
```diff
@@ -69,10 +69,11 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
         level: severity,
       },
       help: {
+        markdown: suggestion.text,
```
michael-go, line 72: there is also `help.text`,
if we set it in `help.markdown` I suggest (🥁 ) we ask the engine to send it as `suggest.markdown` insetead?

src/sarif_converter.ts
```diff
@@ -69,10 +69,11 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
...
         text: suggestion.message,
```
michael-go, line 73: we better remove this line. 
* one reason that [it](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317473) should match the `help.markdown` content - just unformatted ..
* another reason is that `suggestion.message` may be instantiated with values of a specific suggestion instance, containing some symbol names from the code. `rule` should contain the common general data of a rule.

src/sarif_converter.ts
```diff
@@ -60,7 +60,7 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       id: suggestionId,
       name: suggestion.rule,
       shortDescription: {
-        text: suggestion.message,
+        text: suggestion.title || suggestion.message,
```
michael-go, line 63: 🤔 
`suggestion.message` should not be in `rule` - as it can contain values that are specific to a specific instance of a suggestion. So it can make us show unrelated symbol names in the title of of an issue of the same rule.
In addition, `suggestion.message` might be too long sometimes for a title.

To temporarily support rules without `title` maybe do
```
text: suggestion.title  || suggestion.rule
```
instead?

src/sarif_converter.ts
Arvi3d: Since the engine team is going to provide SERIF formatted results, i'd not spend time on talking with them about different names in old proprietary format. Agree?

src/sarif_converter.ts
Arvi3d: suggestion.text != suggestion.message.

suggetion.text will contain only new data in markdown, or empty string.

Is it OK for us to return empty string in markdown here?

src/sarif_converter.ts
Arvi3d: ok

Arvi3d: done

src/sarif_converter.ts
michael-go: > suggestion.text != suggestion.message.

In this comment I'm talking about `help.text`, not `help.markdown`.
`help.text` should be either empty or an unformatted version of `suggestion.text`, but it shouldn't be  `suggestion.message` which should only go to `result` (not to `rule`)

src/sarif_converter.ts
michael-go: do you think that SARIF directly from the engine will happen soon? I worry that it can take much longer until it actually happens. 
Anyway, I wouldn't block this PR because of this - just do think it's an improvement worth suggesting

Arvi3d: thank you. If it was just me, i'd do it without hesitation. But since it requires inter-team communication, i'd leave it as is now.

src/sarif_converter.ts
Arvi3d: done

src/sarif_converter.ts
```diff
@@ -69,10 +65,11 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
         level: severity,
       },
       help: {
-        text: suggestion.message,
+        markdown: suggestion.text,
+        text: '',
       },
       properties: {
```
Spoor2709, line 71: is it possible to use property as a property bag and include a cwe and a language key instead of have them in tags? https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127639

Arvi3d: done

Spoor2709 (approved):
LGTM Great Job!

src/sarif_converter.ts
```diff
@@ -45,22 +45,18 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
...
 
     const suggestionId = suggestion.id;
     const rule = {
       id: suggestionId,
       name: suggestion.rule,
       shortDescription: {
-        text: suggestion.message,
+        text: suggestion.title || suggestion.rule,
       },
       fullDescription: {
```
michael-go, line 61: Sorry can you please remove this (`fullDescription) too? (`suggestion.message` again)

michael-go (approved):
asked for another small semi-related change

src/sarif_converter.ts
Arvi3d: done

PR #35, opened 2020-12-17
fix: stabilise rule id in sarif

miiila:
Remove unnecessary parts and keep the rule id as LANGUAGE/RULE

Ref: https://snyksec.atlassian.net/browse/COD-29

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -50,8 +50,9 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       2: 'warning',
       3: 'error',
     }[suggestion.severity];
-
-    const suggestionId = suggestion.id;
+    // payload comes as URIencoded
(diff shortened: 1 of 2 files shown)


Spoor2709 (approved):
LGTM 😃

src/sarif_converter.ts
```diff
@@ -50,8 +50,9 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
       2: 'warning',
       3: 'error',
     }[suggestion.severity];
-
-    const suggestionId = suggestion.id;
+    // payload comes as URIencoded
+    const language = suggestion.id.split('%2F')[0];
+    const suggestionId = `${language}/${suggestion.rule}`;
```
michael-go, line 55: would rename `suggestionId`  to `ruleId` as otherwise it's quite confusing, as there are multiple suggestions.

PR #41, opened 2021-01-22
[COD-33]Feat: add highlighting to sarif converter

Spoor2709:
[COD-33]

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -102,12 +102,15 @@ const getResults = (suggestions: ISarifSuggestions) => {
   const output = [];
 
   for (const [, suggestion] of <[string, ISarifSuggestion][]>Object.entries(suggestions)) {
-    const result = {
+    let helpers: any[] = [];
+    let result = {
       ruleId: suggestion.id,
       ruleIndex: suggestion.ruleIndex,
       level: suggestion.level ? suggestion.level : 'none',
       message: {
         text: suggestion.text,
+        markdown: suggestion.text,
+        arguments: [''],
       },
       locations: [
         {
@@ -128,17 +131,23 @@ const getResults = (suggestions: ISarifSuggestions) => {
     };
 
     const codeThreadFlows = [];
-    // let i = 0;
+    let i = 0;
     if (suggestion.markers && suggestion.markers.length >= 1) {
       for (const marker of suggestion.markers) {
         for (const position of marker.pos) {
+          const helperIndex = helpers.findIndex(helper => helper.msg === marker.msg);
+          if (helperIndex != -1) {
+            helpers[helperIndex].index.push(i);
+          } else {
+            helpers.push({ index: [i], msg: marker.msg });
+          }
           codeThreadFlows.push({
             location: {
+              id: i,
               physicalLocation: {
                 artifactLocation: {
                   uri: position.file.substring(1),
                   uriBaseId: '%SRCROOT%',
-                  // index: i,
                 },
                 region: {
                   startLine: position.rows[0],
@@ -149,12 +158,13 @@ const getResults = (suggestions: ISarifSuggestions) => {
               },
             },
           });
-          // i += 1;
+          i += 1;
         }
       }
     } else {
       codeThreadFlows.push({
         location: {
+          id: 0,
           physicalLocation: {
             artifactLocation: {
               uri: suggestion.file,
@@ -171,6 +181,11 @@ const getResults = (suggestions: ISarifSuggestions) => {
         },
       });
     }
+
+    const { message, argumentArray } = getArgumentsAndMessage(helpers, result.message.text);
+    result.message.text = message;
+    result.message.arguments = argumentArray;
+
     const newResult = {
       ...result,
       codeFlows: [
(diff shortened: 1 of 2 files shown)


src/sarif_converter.ts
```diff
@@ -128,17 +131,24 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
     if (suggestion.markers && suggestion.markers.length >= 1) {
       for (const marker of suggestion.markers) {
         for (const position of marker.pos) {
+          const helperIndex = helpers.findIndex(helper => helper.msg === marker.msg);
+          if (helperIndex != -1) {
+            helpers[helperIndex].index.push(i);
+          } else {
+            helpers.push({ index: [i], msg: marker.msg });
+          }
+          // result.message.arguments = { ...result.message.arguments, index: { msg: marker.msg } };
```
miiila, line 144: `nit`: leftover

src/sarif_converter.ts
```diff
@@ -149,12 +159,13 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
             },
           });
-          // i += 1;
+          i += 1;
         }
       }
     } else {
       codeThreadFlows.push({
         location: {
+          id: 0,
```
miiila, line 168: `question`: Why is id 0 here? Is this arbitrary value required somewhere else? How shall we make sure consumer is aware of this setup?

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
+    index = str.length + index;
+    if (index < 0) {
+      index = 0;
+    }
+  }
+
+  return str.slice(0, index) + (add || '') + str.slice(index + count);
+}
+
+const getArgumentsAndMessage2 = (
```
miiila, line 220: `required`: Where is `getArgumentsAndMessage1`? :-) Please, change name to something more meaningful. Also, please prefer `function getArgumentsAndMessage2() {}` syntax over variable storing.

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
   }
   return output;
 };
+
+//custom string splice implementation
+function spliceSlice(str: string, index: number, count: number, add: string) {
```
miiila, line 208: `required`: this function can use some refactoring love :-) Couple of suggestions:
- `spliceSlice` is not very clear name
- parameter `add` is consider optional (`add || ''`), but is not marked that way (`add?: string`)
- main difference between `slice` and `splice` on Array is in the in place modification; not sure if this modifies anything, therefore `splice` naming can be misleading
- I somehow believe there might be an easier way how to implement this, but it is hard to say without tests
- this logic is not trivial, therefore it should be tested

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
+
+  return str.slice(0, index) + (add || '') + str.slice(index + count);
+}
+
+const getArgumentsAndMessage2 = (
+  helpers: { index: number[]; msg: number[] }[],
+  message: string,
+): { message: string; arguements: string[] } => {
+  let negativeOffset = 0;
+  let arguements: string[] = [];
```
miiila, line 225: `nit`: typo

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
+  let sortedArguements = helpers.sort((a: any, b: any) => a.msg[0] - b.msg[0]);
+  sortedArguements.forEach((arg: any, index: number) => {
```
miiila, line 227: `optional`: I probably prefer `for (const [index, arg of sortedArguements.entries()) {}`, mainly for readability reasons.

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
+    let word = message.substring(arg.msg[0] + negativeOffset, arg.msg[1] + 1 + negativeOffset);
+    arguements.push(`[${word}]${arg.index.map((i: number) => `(${i})`)}`);
+    message = spliceSlice(
+      message,
+      arg.msg[0] + negativeOffset,
+      arg.msg[1] + 1 + negativeOffset - (arg.msg[0] + negativeOffset),
+      `{${index}}`,
+    );
+    // (2 + index.toString().length) === number of insrted charecters
```
miiila, line 236:
```suggestion
    // (2 + index.toString().length) === number of inserted charecters
```

src/sarif_converter.ts
```diff
@@ -187,3 +203,39 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
+    negativeOffset += arg.msg[0] - (arg.msg[1] + 1) + (2 + index.toString().length);
```
miiila, line 237: `required`: this definitely needs a documentation about reasons for 1 and 2 and for addition vs subtraction

miiila (changes requested):
Great work! Just needs a bit of refactoring and mainly tests. All the changes done are very error prone (mainly off by ones and sign switch), so they need to be tested properly.

src/sarif_converter.ts
Spoor2709: this is the id of if there is no code flow so there is only 1 item giving it an id of 0

src/sarif_converter.ts
Spoor2709: Thank you re the name I completely agree! re the function: Why is this best practise?

miiila: `const getArgumentsAndMessage2 = ()` will end up as an anonymous function in a stack trace (because it is anonymous, just stored in named variable), whilst `function getArgumentsAndMessage2() {}` is named function, so it will appear in stack trace that way.

This is also valid for another features working with function name.

Spoor2709: 👍  thank you

miiila (approved):

PR #47, opened 2021-02-07
[COD-133] feat: add fingerprintings for Sarif output

orkamara:
[COD-133]

This change introduces the new fingerprinting functionality (https://github.com/snyk/octopus/pull/706) as part of the `Code-Client` library as well.

image

[COD-133]: https://snyksec.atlassian.net/browse/COD-133

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -3,6 +3,11 @@ import { Log, ReportingConfiguration, ReportingDescriptor, Result } from 'sarif'
 
 import { IAnalysisResult, IFileSuggestion } from './interfaces/analysis-result.interface';
 
+interface Fingerprint extends IFileSuggestion {
+  version: number;
+  fingerprint: string;
+}
+
 interface ISarifSuggestion extends IFileSuggestion {
   id: string;
   ruleIndex: number;
@@ -10,7 +15,9 @@ interface ISarifSuggestion extends IFileSuggestion {
   level: ReportingConfiguration.level;
   text: string;
   file: string;
+  fingerprints: Fingerprint[];
 }
+
 interface ISarifSuggestions {
   [suggestionIndex: number]: ISarifSuggestion;
 }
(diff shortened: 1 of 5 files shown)


Spoor2709 (approved):
LGTM 👍

src/sarif_converter.ts
```diff
@@ -3,14 +3,21 @@ import { Log, ReportingConfiguration, ReportingDescriptor, Result } from 'sarif'
 
 import { IAnalysisResult, IFileSuggestion } from './interfaces/analysis-result.interface';
 
+interface Fingerprint extends IFileSuggestion {
```
shinin-light, line 6: It doesn't need to extend anything at all.

src/sarif_converter.ts
```diff
@@ -127,9 +134,15 @@ const getResults = (suggestions: ISarifSuggestions) => {
...
           },
         },
-      ],
+      ]
     };
 
+    if (suggestion.fingerprints) {
+      result.fingerprints = {};
+      suggestion.fingerprints.forEach(fingerprinting => {
+        (result.fingerprints as any)[fingerprinting.version] = fingerprinting.fingerprint;
```
shinin-light, line 143: The Sarif format wants fingerprints of type `Record<string, string>`: https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127845
Since the version returned by the API Server is a number at the moment, be sure to wrap it in a string (I know that all the js arrays are basically objects using numbers as keys, but let's be explicit about it and use something like  ``[`${fingerprinting.version}`]``).

shinin-light (approved):

src/sarif_converter.ts
orkamara: Correct! removed it ... :)

src/sarif_converter.ts
orkamara: valid point :) changed...

PR #48, opened 2021-02-09
feat: add sarif support to analyzeFolders

ArturSnyk:
Adding support for sarif on `analyzeFolders` response payload

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -213,6 +213,7 @@ export async function analyzeFolders(
   symlinksEnabled = false,
   maxPayload = MAX_PAYLOAD,
   defaultFileIgnores = IGNORES_DEFAULT,
+  sarif = false,
 ): Promise<IFileBundle> {
   // Get supported filters and test baseURL for correctness and availability
   emitter.supportedFilesLoaded(null);
@@ -279,8 +280,7 @@ export async function analyzeFolders(
     analysisData.analysisResults.files = normalizeResultFiles(analysisData.analysisResults.files, baseDir);
   }
 
-  // Create bundle instance to handle extensions
-  return {
+  const result = {
     baseURL,
     sessionToken,
     includeLint,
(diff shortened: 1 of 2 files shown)


src/analysis.ts
```diff
@@ -213,6 +213,7 @@ export async function analyzeFolders(
   symlinksEnabled = false,
   maxPayload = MAX_PAYLOAD,
   defaultFileIgnores = IGNORES_DEFAULT,
+  sarif = false,
 ): Promise<IFileBundle> {
```
miiila, line 217: `question`: I am surprised we don't have extend an interface here. Is it really there already or there is an opportunity for making it better?

tests/analysis.spec.ts
```diff
@@ -7,104 +8,150 @@ import { sampleProjectPath, bundleFiles, bundleFilesFull } from './constants/sam
...
-    emitter.on(emitter.events.uploadBundleProgress, onUploadBundleProgress);
-
-    // Forse uploading files one more time
-    uploaded = await uploadRemoteBundle(baseURL, sessionToken, bundle.bundleId, bFiles);
-
-    expect(uploaded).toEqual(true);
-
-    expect(onUploadBundleProgress).toHaveBeenCalledTimes(2);
-  }, TEST_TIMEOUT);
+  it(
```
miiila, line 15: `required next-time`: Please split functional and formatting change, it is super hard to recognize what has really changed.

miiila (approved):
🚀

src/analysis.ts
ArturSnyk: it's there, it's 4 levels deep 😅 
```
export interface IBundleResult {
 ...
  readonly sarifResults?: Log;
 ...
}

interface IBundleBase extends IBundleArgs, IBundleResult {}

export interface IGitBundle extends IBundleBase {
...
}

export interface IFileBundle extends IBundleBase {
...
}
```

PR #50, opened 2021-02-11
chore: export AnalysisSeverity interface

ArturSnyk:
we want the consumers of this client to be able
to better type input for client usage


miiila (approved):
There is a typo in commit message (`cunsomers`) + `chore` will not release new version. Which is fine as long as you are aware about that :-)

PR #54, opened 2021-02-17
chore: update lodash to import single functions

Spoor2709:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?
This pr imports lodash single functions instead of the full lodash package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,16 +51,20 @@
     "jsonschema": "^1.2.11"
   },
   "dependencies": {
-    "@types/lodash": "^4.14.167",
+    "@types/lodash.omit": "^4.5.6",
+    "@types/lodash.chunk": "^4.2.6",
+    "@types/lodash.union": "^4.6.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -51,16 +51,20 @@
...
     "@deepcode/dcignore": "^1.0.2",
     "axios": "^0.21.1",
-    "lodash": "^4.17.19",
+    "lodash.omit": "^4.5.0",
+    "lodash.chunk": "^4.2.0",
+    "lodash.union": "^4.6.0",
     "ignore": "^5.1.8",
     "queue": "^6.0.1",
-    "fast-glob": "^3.2.4",
+    "fast-glob": "^3.2.5",
```
ArturSnyk, line 67: btw it won't probably change anything since we're using `^` and that's a patch version upgrade. but worth a try :)

ArturSnyk (approved):

PR #55, opened 2021-02-19
[COD-120] Authentication functionality via Snyk API

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

Authentication functions where rewritten to work with Snyk API

Tests partially cover this case. I did not find a way to test successful workflow automatically

This is braking change for Deepcode integration, as client will not be able to obtain Deepcode API token

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@snyk/code-client",
+  "version": "0.0.1",
   "description": "Typescript consumer of SnykCode public API",
   "main": "dist/index.js",
   "module": "dist/index.es.js",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@snyk/code-client",
+  "version": "0.0.1",
```
miiila, line 3: `required`: please keep version empty - it is automatically handled by semantic release.

miiila (approved):
All good, just remove the version, please.

package.json
Arvi3d: done

PR #60, opened 2021-03-03
[COD-254] fix: wrong number of calculated issues as part of the Sarif conversion

orkamara:
[COD-254]

- [ ] Ready for review

#### What does this PR do?
Fix an issue where ff there are multiple places, where same issue happens, they are lost during conversion to Sarif.

[COD-254]: https://snyksec.atlassian.net/browse/COD-254

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -1,5 +1,5 @@
 // eslint-disable-next-line import/no-unresolved
-import { Log, ReportingConfiguration, ReportingDescriptor, Result } from 'sarif';
+import { Log, ReportingConfiguration, ReportingDescriptor, Result, Tool } from 'sarif';
 
 import { IAnalysisResult, IFileSuggestion, RuleProperties } from './interfaces/analysis-result.interface';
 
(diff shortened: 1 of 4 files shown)


src/sarif_converter.ts
```diff
@@ -96,61 +96,65 @@ const getTools = (analysisResults: IAnalysisResult, suggestions: ISarifSuggestio
...
-    result[suggestionIndex] = {
-      ...suggestions[suggestionIndex],
-      ruleIndex,
-      rule,
-      level: severity,
-      id: suggestionId,
-      text: suggestion.message,
-    };
+    const allIssuesOfSuggestion: ISarifSuggestion[] = allIssuesBySuggestions[suggestionIndex];
+    allIssuesOfSuggestion.forEach(issue => {
```
miiila, line 100: NIT: I would vote more for `.map()` without side effects,
```
allIssues.push(...allIssuesOfSuggestion.map(issue => {...})
```

miiila (approved):
I quickly glanced and found nothing scary. Great job!

Spoor2709 (approved):
LGTM!

src/sarif_converter.ts
orkamara: There are multiple changes I would like to do as part of this file - the code is really complex....

orkamara: As I'm short in time, and this is an important change, I would like us to do more improvements like that soon .... maybe next sprint?

PR #61, opened 2021-03-04
fix: add catergories in a seperate property field in the property bag

Spoor2709:
- [ ] Ready for review

#### What does this PR do?
This PR removes the possibility of having a tag break the category filtering by giving categories it own place in the property bag

This also changes the analysisResults.files[filePath] test to a snapshot due to constant updating of the fingerprinting

### DIFF
--- a/src/sarif_converter.ts
+++ b/src/sarif_converter.ts
@@ -51,9 +51,12 @@ function getSuggestions(analysisResults: IAnalysisResult): ISarifSuggestions {
     }
   }
   return suggestions;
-};
+}
 
-function getRulesAndAllIssues(analysisResults: IAnalysisResult, allIssuesBySuggestions: ISarifSuggestions): { rules: ReportingDescriptor[], allIssues: ISarifSuggestion[] } {
+function getRulesAndAllIssues(
+  analysisResults: IAnalysisResult,
+  allIssuesBySuggestions: ISarifSuggestions,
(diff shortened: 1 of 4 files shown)


orkamara (changes requested):
Still not sure about this change.... it means we don't follow Sarif format anymore.... http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html#_Toc517435925

src/sarif_converter.ts
```diff
@@ -67,10 +70,11 @@ function getRulesAndAllIssues(analysisResults: IAnalysisResult, allIssuesBySugge
     const language = suggestion.id.split('%2F')[0];
     const suggestionId = `${language}/${suggestion.rule}`;
     const ruleProperties: RuleProperties = {
-      tags: [language, ...suggestion.tags, ...suggestion.categories],
+      tags: [language, ...suggestion.tags],
+      categories: suggestion.categories,
```
Spoor2709, line 74: @ork what about:
```suggestion
      tags: [language, ...suggestion.tags, ...categories],
      categories: suggestion.categories,
```

orkamara: I mean, `categories` is not part of Sarif....

Spoor2709: yes but nor is exampleCommitFixes, exampleCommitDescriptions, precision.

orkamara (approved):

PR #62, opened 2021-03-05
fix: flat-cache removed own custom implementation 

Spoor2709:
- [x] Ready for review

#### What does this PR do?
This PR removes the dependency flat-cache that was causing execution issues in the cli due to the export types.
Our implementation only covers the functionality we need and does not support circularJSON

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,0 +1,72 @@
+//This is our own implementation of flat-cache without the use of flattened as we do not need cicular JSON support
+//and the executable for flattened was broken
+import path from 'path';
+import fs from 'fs';
+
+export class Cache {
+  public visited = {};
+  public persisted = {};
+  public pathToFile = '';
+  constructor(docId: string, cacheDir?: any) {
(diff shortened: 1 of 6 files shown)


src/cache.ts
```diff
@@ -0,0 +1,72 @@
...
+    let obj = {};
+
+    let keys = Object.keys(this.visited);
+
+    // no keys visited for either get or set value
+    if (keys.length === 0) {
+      return;
+    }
+
+    keys.forEach(key => {
```
miiila, line 41: `optional`: I would prefer `for..of` and I would refactor to have :
```
for (const key of Object.keys(this.visited)) {}
```

miiila (changes requested):
I pointed out couple of syntactic/design changes I would like to see. From functional point of view, it looks good.

One more thing - since we are introducing new module, we should have it tested. There should be a unit test covering Cache creation, getting, setting and persisting keys + loading existing file.

If we are in hurry, I am OK to skip these test, but only if we test the caching behavior in more integration testing. If skipping please create a jira ticket so we can add tests later.

miiila (approved):

PR #63, opened 2021-03-08
fix: object.fromEntries not supported node10

ArturSnyk:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?
we're using `code-client` on the CLI so we need it to support the minimum version the cli support. which is node 10.
Which doesn't support `object.fromEntries`. This pr uses an internal version of this functionality, that should work the same

This is tested against the same test `object.fromEntries` has on MDN


Spoor2709 (approved):
LGTM!

PR #64, opened 2021-03-16
feat: add support for custom request options

miiila:
- [x] Ready for review
- [x] Linked to [Jira ticket](https://snyksec.atlassian.net/browse/COD-290)

#### What does this PR do?
Adds support for adding custom request options that could be propagated to API server. Headers are currently supported.

As usual, in tests :-)

By complicated manual setup :-)

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -6,6 +6,7 @@ import axios from './axios';
 
 import { IFiles, IFileContent, ISupportedFiles } from './interfaces/files.interface';
 import { IAnalysisResult } from './interfaces/analysis-result.interface';
+import { RequestOptions } from './interfaces/http-options.interface';
 
 type ResultSuccess<T> = { type: 'success'; value: T };
 type ResultError<E> = {
@@ -298,16 +299,19 @@ const CREATE_GIT_BUNDLE_ERROR_MESSAGES: { [P in CreateGitBundleErrorCodes]: stri
   [ErrorCodes.notFound]: 'Unable to found requested repository or commit hash',
 };
 
-export async function createGitBundle(options: {
-  readonly baseURL: string;
-  readonly sessionToken: string;
-  readonly oAuthToken?: string;
-  readonly username?: string;
-  readonly gitUri: string;
-  readonly source: string;
-}): Promise<IResult<RemoteBundle, CreateGitBundleErrorCodes>> {
+export async function createGitBundle(
+  options: {
+    readonly baseURL: string;
+    readonly sessionToken: string;
+    readonly oAuthToken?: string;
+    readonly username?: string;
+    readonly gitUri: string;
+    readonly source: string;
+  },
+  requestOptions?: RequestOptions,
+): Promise<IResult<RemoteBundle, CreateGitBundleErrorCodes>> {
   const { baseURL, sessionToken, oAuthToken, username, gitUri, source } = options;
-  const headers = { 'Session-Token': sessionToken, source };
+  const headers = { ...requestOptions?.headers, 'Session-Token': sessionToken, source };
   if (oAuthToken) {
     headers['X-OAuthToken'] = oAuthToken;
   }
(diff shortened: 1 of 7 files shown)


src/http.ts
```diff
@@ -298,16 +299,19 @@ const CREATE_GIT_BUNDLE_ERROR_MESSAGES: { [P in CreateGitBundleErrorCodes]: stri
...
+    readonly oAuthToken?: string;
+    readonly username?: string;
+    readonly gitUri: string;
+    readonly source: string;
+  },
+  requestOptions?: RequestOptions,
+): Promise<IResult<RemoteBundle, CreateGitBundleErrorCodes>> {
   const { baseURL, sessionToken, oAuthToken, username, gitUri, source } = options;
-  const headers = { 'Session-Token': sessionToken, source };
+  const headers = { ...requestOptions?.headers, 'Session-Token': sessionToken, source };
```
solal-snyk, line 314: I see `requestOptions` is an optional, what happens if they're missing?

tests/git.analysis.spec.ts
```diff
@@ -191,6 +192,39 @@ describe('Functional test of analysis', () => {
...
+    'passes custom options correctly',
+    async () => {
+      try {
+        await analyzeGit(
+          {
+            baseURL,
+            sessionToken,
+            includeLint: false,
+            severity: 1,
+            gitUri: 'git@github.com:DeepCodeAI/cli.git',
```
solal-snyk, line 211: This makes the test reliant on the network (which means we're ultimately testing the network too) - given what we're trying to achieve here, I guess there's no other way and we can't mock this?

miiila: @solal-snyk This is the first test, that is mocking network here 😂 We have a lot of network dependent tests there. This one is not using network because of https://github.com/snyk/code-client/pull/64/files/e46f2e315e62fa0da97a7655c2b2515993caa028#diff-470745133bb7400e74c85a9f84a74d1494a4ec040d6d0908e617f8d217132209R198

axios is mocked to reject request immediately, so network doesn't play any role here.

src/http.ts
miiila: Nothing :-) Question mark is checking for it being available: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

tests/git.analysis.spec.ts
solal-snyk: Awesome thanks for explaining just learned something very useful regarding our stack 😄

src/http.ts
solal-snyk: Sounds good, thanks for sharing the docs 😀

solal-snyk (approved):

tests/api.spec.ts
```diff
@@ -367,7 +324,7 @@ describe('Requests to public API', () => {
 
       if (response.value.status === AnalysisStatus.done) {
         expect(response.value.analysisURL.includes(realBundleIdFull)).toBeTruthy();
-        expect(Object.keys(response.value.analysisResults.suggestions).length).toEqual(8);
+        expect(Object.keys(response.value.analysisResults.suggestions).length).toEqual(6);
```
orkamara, line 327: question: why did we change this file?

src/http.ts
```diff
@@ -434,7 +441,7 @@ export async function getAnalysis(options: {
   // ?linters=false is still a truthy query value, if(includeLint === false) we have to avoid sending the value altogether
   const params = { severity, linters: includeLint || undefined };
 
-  const headers = { 'Session-Token': sessionToken, source };
+  const headers = { ...requestOptions?.headers, 'Session-Token': sessionToken, source };
```
orkamara, line 444: question: maybe worth using white list here? for all the headers we would like to send..? (thinking about future confusions of passing unnecessary data as part of the request options...

tests/api.spec.ts
miiila: @orkamara Because tests were failing on master. Check https://snyk.slack.com/archives/C0135Q8LP61/p1615978344024300

src/http.ts
miiila: @orkamara I am not sure what do you mean. As we are in TypeScript, current type is supporting `headers` key only. So all you can do is to add headers. If you would like to add something else, you would need to extend the type first.

PR #65, opened 2021-03-18
feat: axios agent keep alive to reuse connections

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

Under heavy load i realized, that we don't reuse connections and establish new one al the time. At some point, we just can't make a new connection and everything hangs.

This PR aims to solve this by setting special http/https agent with keep alive and granular socket settings

### DIFF
--- a/src/axios.ts
+++ b/src/axios.ts
@@ -7,7 +7,6 @@ const agentOptions = {
   keepAlive: true,
   maxSockets: 100, // Maximum number of sockets to allow per host. Defaults to Infinity.
   maxFreeSockets: 10,
-  // timeout: 60000, // active socket keepalive for 60 seconds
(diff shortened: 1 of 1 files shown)


src/axios.ts
```diff
@@ -1,11 +1,25 @@
 import axios, { AxiosRequestConfig, AxiosError, AxiosResponse } from 'axios';
+import http from 'http';
+import https from 'https';
 import emitter from './emitter';
 
+const agentOptions = {
+  keepAlive: true,
+  maxSockets: 100, // Maximum number of sockets to allow per host. Defaults to Infinity.
+  maxFreeSockets: 10,
+  // timeout: 60000, // active socket keepalive for 60 seconds
```
miiila, line 10: `question`: should this be commented out?

miiila (approved):

src/axios.ts
Arvi3d: done

PR #76, opened 2021-04-19
[COD-360] fix: fixed parsing of negative ignores and added tests

shinin-light:
[COD-360]
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

Our current `.gitignore` parsing doesn't work with negative ignores (!-rules). Long story short, it seems that the library on which we base our glob processing (https://github.com/micromatch/micromatch) is not working as expected.

For example, an option we use called `basename` is described as "If set, then patterns without slashes will be matched against the basename of the path if it contains slashes", but in reality it always compares the pattern to the basename of the file, regardless if the pattern has a slash or not. This specific feature was already addressed [by someone who wanted to use micromatch to parse `.gitignore` files](https://github.com/micromatch/micromatch/issues/83), and it was [briefly fixed in version 3.x](https://github.com/micromatch/micromatch/pull/76/commits/226797c336dd64d839dd7f5e0043fbed5237dd2f#diff-6eafe9bf073592b4625339b25a7714193c251bf86758a47b1f9c93e649f27e81), however they since reverted the change and [now there is no mention of the tests](https://github.com/micromatch/micromatch/blob/master/test/issue-related.js) that should have future-proofed this feature.
Another issue is that positive glob matching (i.e. finding all the files matching our filters) and negative glob matching (i.e. excluding files following the .gitignore rules) cannot be unified in a single call to micromatch. Trying to use intersections with negations-of-negations would not work properly due to Set Theory. Therefore we need to first filter based on extension and then again based on ignores.

Check implementation of `filterIgnoredFiles` and `parseFileIgnores` and their occurrences.

- have a running deepcode instance locally
- `SNYK_API_HOST=http://localhost:8080 SNYK_API_KEY=<a deepcode sessionToken from your GitHub account in the snyk org> npm run test`

[COD-360]: https://snyksec.atlassian.net/browse/COD-360


src/files.ts
```diff
@@ -132,25 +135,22 @@ export function determineBaseDir(paths: string[]): string {
   return '';
 }
 
-function searchFiles(
+async function* searchFiles(
```
Spoor2709, line 138: I learn something new everyday function* cool thanks 😃

shinin-light: It'a generator: instead of returning a value or a promise itself, it returns an async generator so you can then use `for (await value of results)...`

Spoor2709 (approved):
This really took me some time to get my head around! Great Job it looks good to me!

miiila (approved):

Arvi3d (approved):

PR #78, opened 2021-04-30
Fix: Propagate --insecure parameter

miiila:
To unify behaviour with default `snyk test` command, this PR takes an advantage of https://github.com/snyk/snyk/commit/5ae3182e16929b0d6ac87a976194c599b22835e3#diff-1bc78dd2ada161f97e1a1e595957fbe2487468b6a81191600acf0aa63769ad88R136

### DIFF
--- a/src/axios.ts
+++ b/src/axios.ts
@@ -3,12 +3,21 @@ import http from 'http';
 import https from 'https';
 import emitter from './emitter';
 
+// Snyk CLI allow passing --insecure flag which allows self-signed certificates
+// It updates global namespace property ignoreUnknownCA and we can use it in order
+// to pass rejectUnauthorized option to https agent
+export declare interface Global extends NodeJS.Global {
+  ignoreUnknownCA: boolean;
+}
+declare const global: Global;
+
 const agentOptions = {
   keepAlive: true,
   maxSockets: 100, // Maximum number of sockets to allow per host. Defaults to Infinity.
   maxFreeSockets: 10,
   freeSocketTimeout: 60000, // // Maximum number of sockets to leave open for 60 seconds in a free state. Only relevant if keepAlive is set to true. Defaults to 256.
   socketActiveTTL: 1000 * 60 * 10,
+  rejectUnauthorized: !global.ignoreUnknownCA,
 };
 
 const axios_ = axios.create({
--- a/tests/api.spec.ts
+++ b/tests/api.spec.ts
@@ -42,7 +42,7 @@ describe('Requests to public API', () => {
     if (response.type === 'error') return;
     expect(new Set(response.value.configFiles)).toEqual(new Set(['.dcignore', '.gitignore']));
     expect(new Set(response.value.extensions)).toEqual(
-      new Set(['.es', '.es6', '.htm', '.html', '.js', '.jsx', '.py', '.ts', '.tsx', '.vue', '.java']),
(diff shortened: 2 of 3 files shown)


src/axios.ts
```diff
@@ -3,12 +3,19 @@ import http from 'http';
...
+export declare interface Global extends NodeJS.Global {
+  ignoreUnknownCA: boolean;
+}
+declare const global: Global;
+
 const agentOptions = {
   keepAlive: true,
   maxSockets: 100, // Maximum number of sockets to allow per host. Defaults to Infinity.
   maxFreeSockets: 10,
   freeSocketTimeout: 60000, // // Maximum number of sockets to leave open for 60 seconds in a free state. Only relevant if keepAlive is set to true. Defaults to 256.
```
miiila, line 16: @Arvi3d I am fairly sure this settings does nothing here 😇 It is not supported at https://nodejs.org/docs/latest-v12.x/api/http.html#http_new_agent_options and it looks like it is used only, when https://www.npmjs.com/package/agentkeepalive is around. I'd vote for removing it and typing agentOptions properly.

solal-snyk (approved):

src/axios.ts
```diff
@@ -3,12 +3,19 @@ import http from 'http';
 import https from 'https';
 import emitter from './emitter';
 
+// For a compatibility with Snyk CLI
+export declare interface Global extends NodeJS.Global {
+  ignoreUnknownCA: boolean;
```
Spoor2709, line 8: Required: Does CA stand for certification Authorities? Could we add a comment explaining that?

src/axios.ts
```diff
@@ -3,12 +3,19 @@ import http from 'http';
 import https from 'https';
 import emitter from './emitter';
 
+// For a compatibility with Snyk CLI
+export declare interface Global extends NodeJS.Global {
```
Spoor2709, line 7: Question: maybe I am lacking some understanding but why do you have to extend the global object? Is this because typescript does not define the property that is already there?

tests/api.spec.ts
```diff
@@ -42,7 +42,7 @@ describe('Requests to public API', () => {
     if (response.type === 'error') return;
     expect(new Set(response.value.configFiles)).toEqual(new Set(['.dcignore', '.gitignore']));
     expect(new Set(response.value.extensions)).toEqual(
-      new Set(['.es', '.es6', '.htm', '.html', '.js', '.jsx', '.py', '.ts', '.tsx', '.vue', '.java']),
+      new Set(['.es', '.es6', '.htm', '.html', '.js', '.jsx', '.py', '.ts', '.tsx', '.vue', '.java', '.java-dummy']),
```
Spoor2709, line 45: Question: Why do we now return .java-dummy in this list?

Spoor2709 (changes requested):
Great Job this looks great, 1 super small change requested

tests/api.spec.ts
miiila: Ask t-42 ¯\_(ツ)_/¯ Looks like something reachability related.

src/axios.ts
miiila: Global is an Interface with defined properties and TypeScript interfaces are closed for additions. So we need to create our own interface.

src/axios.ts
miiila: `CA` should be widely known abbrev, so I don't wanna to add a comment just because of that. However, I can elaborate on a comment explaining, how it can be turned on.

shinin-light (approved):

Spoor2709 (approved):

PR #81, opened 2021-05-11
COD-409/add reachability args

pkey:
- [x] Ready for review
- [x] Linked to Jira ticket
- [x] Figure out failing tests

#### What does this PR do?

- Adds missing reachability args to enable reachability via `analyzeGit`
- Adds additional tests

PR #82, opened 2021-05-12
Cod 409/reachability fix tests

pkey:
#### What does this PR do?

Adds a tests to test reachability following the introduced missing args in https://github.com/snyk/code-client/pull/81. 
The reason these were not there before is that new `suggestion` image was not yet deployed and it is needed in order to successfully run the tests.

### DIFF
--- a/tests/git.analysis.spec.ts
+++ b/tests/git.analysis.spec.ts
@@ -182,6 +182,22 @@ describe('Functional test of analysis', () => {
       TEST_TIMEOUT,
     );
 
+    it(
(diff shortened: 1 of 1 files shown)


miiila:
@pkey please keep branch prefixes one of `fix/chore/feat` etc. Not a hard requirement, but very nice to keep consistency :-)

miiila (approved):

neonnoon (approved):
Looks good to me, as far as I can judge from my limited context.
It's always great to see a test being red first, going to green.

PR #85, opened 2021-05-21
fix: fixed sarif converter filepath trimming

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket


shinin-light (approved):
I'll approve this. However afaik the current API Server always returns beginning slashes, while the new API should never return it (more specifically, it returns exactly what you put in it without manipulating it, so uploaded bundles may have slashed paths if they are uploaded as such)

PR #86, opened 2021-05-25
feat: expose bundle creation logic

pkey:
- [x] Write separately for the new method. It should be covered by `analyzeFolders` tests.

#### What does this PR do?

Exposes `createBundleFromFolders` method which creates a bundle remotely and returns `bundleId`. The logic is extracted and reused (with some caveats) in the `analyzeFolders` method.

\-
\-

This is being done to allow us to separate logic which uploads and which analyses so that we can perform this in separate steps.

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -46,6 +46,19 @@ import { ISupportedFiles } from './interfaces/files.interface';
 
 const sleep = (duration: number) => new Promise(resolve => setTimeout(resolve, duration));
 
+const ANALYSIS_OPTIONS_DEFAULTS = {
+  baseURL: defaultBaseURL,
+  sessionToken: '',
+  includeLint: false,
+  reachability: false,
+  severity: AnalysisSeverity.info,
+  symlinksEnabled: false,
+  maxPayload: MAX_PAYLOAD,
+  defaultFileIgnores: IGNORES_DEFAULT,
+  sarif: false,
+  source: '',
+}
+
 async function pollAnalysis(
   {
     baseURL,
@@ -239,20 +252,9 @@ function mergeBundleResults(bundle: IFileBundle, analysisData: IBundleResult, li
     analysisResults,
   };
 }
-let analyzeFolderDefaults = {
-  baseURL: defaultBaseURL,
-  sessionToken: '',
-  includeLint: false,
-  reachability: false,
-  severity: AnalysisSeverity.info,
-  symlinksEnabled: false,
-  maxPayload: MAX_PAYLOAD,
-  defaultFileIgnores: IGNORES_DEFAULT,
-  sarif: false,
-  source: '',
-};
+
 export async function analyzeFolders(options: FolderOptions): Promise<IFileBundle> {
-  const analysisOptions: AnalyzeFoldersOptions = { ...analyzeFolderDefaults, ...options };
+  const analysisOptions: AnalyzeFoldersOptions = { ...ANALYSIS_OPTIONS_DEFAULTS, ...options };
   const {
     baseURL,
     sessionToken,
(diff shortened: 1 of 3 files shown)


src/analysis.ts
```diff
@@ -482,3 +446,81 @@ export async function analyzeGit(options: GitOptions, requestOptions?: RequestOp
 
   return result;
 }
+
+interface CreateBundleFromFolders extends FolderOptions {
```
novalex, line 450: nit/optional: not an ideal or descriptive enough name IMO, should be something like `CreateBundleFromFoldersOptions`.

src/analysis.ts
```diff
@@ -482,3 +446,81 @@ export async function analyzeGit(options: GitOptions, requestOptions?: RequestOp
...
+}
+
+let createBundleFromFoldersDefaults = {
+  baseURL: defaultBaseURL,
+  sessionToken: '',
+  symlinksEnabled: false,
+  maxPayload: MAX_PAYLOAD,
+  defaultFileIgnores: IGNORES_DEFAULT,
+  source: '',
+};
```
novalex, line 463: nit: I would put these defaults inline on L472, since the var is not used anywhere else I don't see the point of declaring it separately in the file scope.

novalex (commented):
Nice refactor overall 💯

shinin-light (approved):

tests/analysis.spec.ts
```diff
@@ -12,156 +12,188 @@ import { AnalysisSeverity } from '../src/interfaces/analysis-result.interface';
...
-        }
-      });
-      emitter.on(emitter.events.supportedFilesLoaded, onSupportedFilesLoaded);
+  describe('analyzeFolders', () => {
+    it(
+      'analyze folder',
+      async () => {
+        const onSupportedFilesLoaded = jest.fn((data: ISupportedFiles | null) => {
+          if (data === null) {
+            // all good
```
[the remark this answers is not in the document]
pkey: @shinin-light No idea tbh, this one was here before.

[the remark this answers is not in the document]
pkey: [stepsize] Investigate possibly redundant test

src/analysis.ts
novalex: It's still outside, but now a const, I meant just putting those properties directly in the method where they're used, like:
```
export async function createBundleFromFolders(options: CreateBundleFromFolders): Promise<RemoteBundle | null> {
  const analysisOptions = {
    baseURL: defaultBaseURL,
    sessionToken: '',
    symlinksEnabled: false,
    maxPayload: MAX_PAYLOAD,
    defaultFileIgnores: IGNORES_DEFAULT,
    source: '',
    ...options,
  };

novalex:
> moved defaults to a separate file for all of the methods.

LIES! 😄 

Usually I try to always put the data closest to where it's used - having a variable at the top of the file, and using it in a single method at the bottom introduces additional context switching.

Non-blocking though, just cleaner IMO.

novalex (approved):

pkey:
> > moved defaults to a separate file for all of the methods.
> 
> LIES! 😄
> 
> Usually I try to always put the data closest to where it's used - having a variable at the top of the file, and using it in a single method at the bottom introduces additional context switching.
> 
> Non-blocking though, just cleaner IMO.

@novalex In this case, this variable is being shared between multiple methods :p

src/analysis.ts
pkey: As I mentioned, I have extracted it and reused in multiple methods as I noticed same things are being repeated.

pkey:
Cool, tested it with the new service and it works 🎉 merging!

PR #87, opened 2021-05-25
[COD-454] Feat: loosening up ignore correctness to improve performances

shinin-light:
[COD-454]
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
This PR improves performances at the cost of loosening up the ignore correctness a bit, without however completely neglecting negative ignores like before.
In the current version (prior to this PR) we initially filter only by the extension patterns returned by the backend, and then for each file we ask our glob matcher if that file should be ignored or not. This leads to exploring every single branch in the file system (e.g. `node_modules/**`) and to check for each file that it should be ignored before proceeding to the next. This is extremely inefficient as the majority of the branches end up ignored anyway. Also, while exploring the ignored branches, the library "looks stuck" because the updates are emitted only if a valid file is found.

The idea coming from @Arvi3d is that in most cases, negative ignores are just used to exclude a path that was ignored previously, and the length of these exception chains is usually just 2 (even if they could be much longer). Therefore, by manually exploring the negative ignore cases after the main bulk is computed, we should be able to address most of the cases at a reasonable speed.

In this PR:
- Split `searchFiles` so that it uses the positive ignores directly to prune the ignored branches of the file system, and then searches the negative ignores paths and matches them against the backend filters.
- Normalyzed the matcher libraries so that just use `fast-glob` for FS search and `multimatch` for manual glob matching everywhere.

Technical Debt: the following `.gitignore` edge case still fails to be correctly addressed (as stated in the tests):
`node_module/` <= ignores everything in a `node_module` folder and it's relative subfolders
`!node_module/my_module/` <= excludes the `my_module` subfolder from the ignore
`node_module/my_module/build/` <= re-includes the `build` subfolder in the ignore
we should explore the possibility to fork the `globby` library and extend the ignore files (it only natively supports `.gitignore` files, not `.dcignore`) to use it directly instead of parsing the ignore files ourselves and then use `fast-glob` and `multimatch`.

#### Additional questions

[COD-454]: https://snyksec.atlassian.net/browse/COD-454


src/files.ts
```diff
@@ -39,7 +38,7 @@ export function notEmpty<T>(value: T | null | undefined): value is T {
   return value !== null && value !== undefined;
 }
 
-const microMatchOptions = { basename: true, dot: true, posixSlashes: true };
+const miniMatchOptions = { matchBase: true, dot: true };
```
pkey, line 41: optional: seems like this should be `multiMatchOptions`

tests/api.spec.ts
```diff
@@ -107,15 +120,18 @@ describe('Requests to public API', () => {
...
-        `/GitHubAccessTokenScrambler12.java`,
-        `/app.js`,
-        `/db.js`,
-        `/main.js`,
-        `/not/ignored/this_should_not_be_ignored.java`,
-        `/routes/index.js`,
-        `/routes/sharks.js`,
-      ]);
+      expect(response.value.missingFiles).toEqual(fakeMissingFiles);
+      //   [
```
pkey, line 124: required: comments :p

tests/bigrepo.js
```diff
@@ -0,0 +1,34 @@
+const codeClient = require('../dist/index');
```
pkey, line 1: required: I believe this should not be committed?

On the other hand - shall we have an actual performance tests here? I think we learned a lesson that we don't want to accidentally end up in the same situation again.

src/files.ts
```diff
@@ -65,7 +63,7 @@ export function parseFileIgnores(path: string): string[] {
 
     rules = f
       .split('\n')
-      .map(l => l.trim().replace(/\/$/, '')) // Remove white spaces and trim slashes
```
pkey, line 68: question: why don't we need this anymore?

src/files.ts
```diff
@@ -141,14 +153,37 @@ async function* searchFiles(
...
-      yield filePath;
+    yield filePath;
+  }
+  // TODO: This is incorrect because the .gitignore format allows to specify exceptions to previous rules, therefore
+  // the separation between positive and negative ignores is incorrect in a scenario with 2+ exeptions like the one below:
+  // `node_module/` <= ignores everything in a `node_module` folder and it's relative subfolders
+  // `!node_module/my_module/` <= excludes the `my_module` subfolder from the ignore
+  // `node_module/my_module/build/` <= re-includes the `build` subfolder in the ignore
+  if (negativeIgnores.length) {
+    const searcher2 = fg.stream(negativeIgnores, {
```
pkey, line 173: required: `searcher2` :D I suggest splitting this into separate function to make it a little bit more readable.

src/files.ts
```diff
@@ -76,24 +74,38 @@ export function parseFileIgnores(path: string): string[] {
...
+      // eslint-disable-next-line no-param-reassign
       rule = rule.substring(1);
-      prefix = "!";
+      prefix = '!';
     }
-    return rule.startsWith('/') || rule.startsWith('**')
-      ? prefix + nodePath.posix.join(dirname, rule, '**')
-      : prefix + nodePath.posix.join(dirname, '**', rule, '**');
-  });
+    // Mappings from .gitignore format to glob format:
```
pkey, line 84: optional: I have spent quite some time to try and understand what's going on. I think the mapping would benefit from being split into a separate function.

tests/bigrepo.js
pkey: Want to stress the performance tests. I would imagine it will take some extra effort but could be beneficial for us to be more confident in our changes in the future.

pkey:
@Arvi3d @shinin-light In general, very cool approach! Nice idea and nice execution.

src/files.ts
shinin-light: multimatch uses `minimatch` underneath and uses minimatch options structure. We could rename it but I don't think it's needed.

tests/bigrepo.js
shinin-light: I thought I've removed this file from the commit. I think we should have a performance test instead (but I'm not sure how to test it properly yet)

tests/analysis.spec.ts
```diff
@@ -84,13 +84,18 @@ describe('Functional test of analysis', () => {
             isSupported: true,
             lang: 'JavaScript',
           },
+          {
+            files: 1,
```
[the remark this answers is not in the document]
shinin-light: yeah :(

src/files.ts
shinin-light: We don't want to remove the trailing slashes at this point as it messes up with our ability later on to decide the correct parsing from gitignore format to glob.

tests/api.spec.ts
```diff
@@ -16,10 +16,23 @@ import {
...
+const realBundleId = '1d16a90ae18e96edf2ab21e442a7c0665939613e4a2a9831f98c7cf32bcd5e00';
 let realBundleIdFull = '';
+const extendedBundleId = '587a6bcb0095606ad57ccc7bb7ac6401475ce4181c13f7136491a16df06544f1';
+
+const fakeMissingFiles = [
+  `/GitHubAccessTokenScrambler12.java`,
+  `/app.js`,
+  `/db.js`,
+  `/main.js`,
+  // TODO: This should be ignored
```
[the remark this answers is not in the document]
shinin-light: Hard to do since we use the same sample repo to do all the folder tests. An option would be to have multiple sample repos each one for a given test case but I think it would be hard to maintain.

src/files.ts
```diff
@@ -141,14 +165,44 @@ async function* searchFiles(
...
     followSymbolicLinks: symlinksEnabled,
+    ignore: positiveIgnores,
   });
-  for await (const filePath of searcher) {
-    if (filterIgnoredFiles([filePath.toString()], ignores).length) {
-      yield filePath;
+  for await (const filePath of positiveSearcher) {
+    yield filePath;
+  }
+  // TODO: This is incorrect because the .gitignore format allows to specify exceptions to previous rules, therefore
```
pkey, line 186: Tech debt?

tests/constants/sample.ts
```diff
@@ -27,14 +31,19 @@ export const bundleFilePaths = [
   'main.js',
   'routes/index.js',
   'routes/sharks.js',
+  // TODO: This should be ignored for consistency with the .gitignore format (see last rule above),
```
pkey, line 34: Tech debt?

pkey (approved):
Looks good otherwise and I think we either should write a performance test now or add it as a tech debt.

Reason being that if we touch this part again (and we probably will, provided that the solution has know issues) - we will have to have some sort of performance metric anyways.

PR #89, opened 2021-06-08
[NEB-13]feat: align message.text to github format

ArturSnyk:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

When using the sarif output on github, we're failing to parse the `message.text` field. as we use `message.text` <-> `message.markdown` the opposite way github is using it.
We would like to align those two options. as the [sarif ](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317472) indicates that the `argument` property is used by both `text` and `markdown` it won't change any logic if we use one or the other.


src/sarif_converter.ts
```diff
@@ -209,7 +209,7 @@ function getResults(allIssues: ISarifSuggestion[]): Result[] {
 
     if (result.message.text) {
```
pkey, line 210: question: same question as in [previous PR](https://github.com/snyk/octopus/pull/911#discussion_r651865003), is the predicate correct?

pkey:
Seems like we have the same logic spread out between multiple codebases - probably worth considering extracting them and reusing?

pkey (approved):

PR #93, opened 2021-07-02
[NEB-44] fix: fixed result merge for extended analysis

shinin-light:
[NEB-44]
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

Added the result merge for extended analyses by parsing the sarif results.
All input (i.e. `extendAnalysisOptions.files`) and output (i.e. sarif positions) are always to be considered relative to `baseDir` now for consistency.

[NEB-44]: https://snyksec.atlassian.net/browse/NEB-44

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -147,15 +147,42 @@ function mergeBundleResults(
   const sarifRules = newAnalysisResults.sarif.runs[0].tool.driver.rules || [];
   const oldResults = oldAnalysisResults.sarif.runs[0].results || [];
   for (let res of oldResults) {
-    const locations: string[] = (res.locations || []).flatMap(
-      loc => !!loc.physicalLocation?.artifactLocation?.uri ? [loc.physicalLocation.artifactLocation.uri] : []
+    // NOTE: Node 10 doesn't support the more readable .flatMap, so we need to use .reduce, but the behaviour would be the following:
+    // const locations: string[] = (res.locations || []).flatMap(
+    //   loc => !!loc.physicalLocation?.artifactLocation?.uri ? [loc.physicalLocation.artifactLocation.uri] : []
+    // );
+    // const codeFlowLocations: string[] = (res.codeFlows || []).flatMap(
+    //   cf => (cf.threadFlows || []).flatMap(
+    //     tf => (tf.locations || []).flatMap(
+    //       loc => !!loc.location?.physicalLocation?.artifactLocation?.uri ? [loc.location.physicalLocation.artifactLocation.uri] : []
+    //     )
+    //   )
+    // );
+    const locations: string[] = (res.locations || []).reduce<string[]>(
+      (acc, loc) => {
+        if (loc.physicalLocation?.artifactLocation?.uri) {
+          acc.push(loc.physicalLocation!.artifactLocation!.uri!);
+        };
+        return acc;
+      }, []
     );
-    const codeFlowLocations: string[] = (res.codeFlows || []).flatMap(
-      cf => (cf.threadFlows || []).flatMap(
-        tf => (tf.locations || []).flatMap(
-          loc => !!loc.location?.physicalLocation?.artifactLocation?.uri ? [loc.location.physicalLocation.artifactLocation.uri] : []
-        )
-      )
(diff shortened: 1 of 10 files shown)


src/analysis.ts
```diff
@@ -127,36 +127,73 @@ export async function analyzeFolders(options: FileAnalysisOptions): Promise<File
...
+  newAnalysisResults: AnalysisResult,
+  limitToFiles: string[],
+  removedFiles: string[] = [],
+): AnalysisResult {
+  // Start from the new analysis results
+  // For each finding of the old analysis,
+  //  if it's location is not part of the limitToFiles or removedFiles (removedFiles should also be checked against condeFlow),
+  //   append the finding to the new analysis and check if the rule must be added as well
+  const changedFiles = [...limitToFiles, ...removedFiles];
+  const sarifResults = (newAnalysisResults.sarif.runs[0].results || []).filter(res => {
```
neonnoon, line 141: What's the meaning of `newAnalysisResults.sarif.runs[0]`? Specifically reading only the first element of an array kind of raises a flag..

src/analysis.ts
```diff
@@ -127,36 +127,73 @@ export async function analyzeFolders(options: FileAnalysisOptions): Promise<File
...
+      // TODO: This should not be necessary in theory but, in case of two identical files,
+      // Bundle Server returns the finding in both files even if limitToFiles only reports one
+      const loc = res.locations && res.locations[0].physicalLocation?.artifactLocation?.uri;
```
neonnoon, line 144: Just fyi, you can also use optional chaining with arrays `res.locations?.[0]`.

src/analysis.ts
```diff
@@ -127,36 +127,73 @@ export async function analyzeFolders(options: FileAnalysisOptions): Promise<File
...
+      return loc && changedFiles.includes(loc);
+  });
+  const sarifRules = newAnalysisResults.sarif.runs[0].tool.driver.rules || [];
+  const oldResults = oldAnalysisResults.sarif.runs[0].results || [];
+  for (let res of oldResults) {
+    const locations: string[] = (res.locations || []).reduce<string[]>(
```
neonnoon, line 150: Question: Aren't these `reduce`s basically `map().filter()` / `flatMap().filter()`?
No need to change, but might be easier to read.

src/files.ts
```diff
@@ -295,8 +295,10 @@ export async function prepareExtendingBundle(
   let processingFiles: string[] = filterSupportedFiles(files, supportedFiles);
 
   // Exclude files to be ignored based on ignore rules. We assume here, that ignore rules have not been changed.
-  processingFiles = processingFiles.filter(f => !isMatch(f, fileIgnores));
-
+  processingFiles = processingFiles
+    .map(f => resolveBundleFilePath(baseDir, f))
+    .filter(f => !isMatch(resolveBundleFilePath(baseDir, f), fileIgnores));
```
neonnoon, line 300: This seems weird, the map above maps `processingFiles` with `resolveBundleFilePath`. Why do you need to pass it here again.

Put differently, what is `resolveBundleFilePath(baseDir, resolveBundleFilePath(baseDir, f))`?

neonnoon (approved):
Approving, as most comments are cosmetic or minor, but I'd appreciate if you could have a look at them.

src/analysis.ts
shinin-light: We only report 1 analysis result per SARIF (the format allows you to report multiple analyses at once)

src/analysis.ts
shinin-light: Great suggestion, It didn't occur to me to simplify it that way

src/analysis.ts
shinin-light: Didn't know, updating

src/files.ts
shinin-light: It should be idempotent (from testing it), but removed for readability.

src/analysis.ts
shinin-light: Bad news, we need to be compatible with Node 10 and flatMap has been introduced in Node 11... I'm reverting the changes.

Spoor2709 (approved):
Looks Good to me if you want to comment above what the reduces are doing, I found it semi confusing at first look 😃

PR #95, opened 2021-07-23
Fix: migrated axios needle

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

Migrated axios needle

Removed unsupported linter argument

Thanks to this change, our CLI will not need to depend on axios in addition to needle


README.md
```diff
@@ -121,7 +118,7 @@ const result = await codeClient.extendBundle({
   },
   removedFiles: [],
 });
-const { bundleId, missingFiles, uploadURL } = result;
```
ArturSnyk, line 124: why are we removing the `uploadURL` attribute from docs?

src/analysis.ts
```diff
@@ -258,7 +250,6 @@ export async function analyzeFolders(options: FolderOptions): Promise<IFileBundl
   const {
     baseURL,
     sessionToken,
-    includeLint,
```
ArturSnyk, line 261: Since we're using analyze folder on the CLI,  although we don't send includeLint there, we do have some mock response we compare it with. where the response payload do have this property https://github.com/snyk/snyk/blob/master/test/fixtures/sast/sample-analyze-folders-response.json#L4
this raise a question. would do to this PR or/and similar changes on the new API will change the payload we send upon analyzing?

README.md
Arvi3d: it was deprecated and not used by anyone for a very long time. So it's just a clean-up

src/analysis.ts
Arvi3d: yes, i'll work on the CLI support separately. 
But this is a valid point right now, to make sure, we don't break current CLI tests. I'll do it.

ArturSnyk (approved):

src/analysis.ts
Arvi3d: So, CLI depends on a specific version of code-client, so I can assume, we are not going to break CLI tests by introducing new version here. I'm going to merge this PR and then test CLI with the newly created version. Will make a separate PR there.

package.json
```diff
@@ -61,12 +62,13 @@
     "@types/lodash.union": "^4.6.6",
     "@types/sarif": "^2.1.3",
     "@types/uuid": "^8.3.0",
-    "axios": "^0.21.1",
     "ignore": "^5.1.8",
+    "ipaddr.js": "^2.0.1",
```
michelkaporin, line 66: You can remove `ipaddr.js` as it's not needed for getIpFamily() function any more.

Arvi3d: done

src/needle.ts
```diff
@@ -0,0 +1,77 @@
...
+    family: payload.family,
+    json: true,
+    compressed: true, // sets 'Accept-Encoding' to 'gzip, deflate, br'
+    follow_max: 5, // follow up to five redirects
+    rejectUnauthorized: !global.ignoreUnknownCA, // verify SSL certificate
+    open_timeout: 0,
+    agent,
+  };
+
+  emitter.apiRequestLog(`=> HTTP ${method?.toUpperCase()} ${url} ${data ?? ''}`);
```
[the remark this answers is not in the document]
Arvi3d: done

michelkaporin (approved):
Looks good, thank you! 🙌

PR #97, opened 2021-07-28
New code api support

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

This PR finally is about to publish support for the new Code API.
It introduces breaking changes in interfaces.


package.json
```diff
@@ -31,7 +31,16 @@
   "homepage": "https://github.com/snyk/code-client#readme",
   "jest": {
     "preset": "ts-jest",
-    "testEnvironment": "node"
+    "testEnvironment": "node",
+    "testTimeout": 50000,
```
pkey, line 35: question: any reason to increase the timeout here?

src/interfaces/analysis-result.interface.ts
```diff
@@ -1,123 +1,19 @@
...
-
-export interface IFileBundle extends IBundleBase {
-  readonly baseDir: string;
-  readonly paths: string[];
-  readonly supportedFiles: ISupportedFiles;
-  readonly fileIgnores: string[];
-  readonly symlinksEnabled: boolean;
+  coverage: Coverage[];
+  status: 'COMPLETE';
+  // progress: 1;
```
pkey, line 18: probably not needed

src/files.ts
```diff
@@ -266,21 +283,23 @@ export async function* collectBundleFiles(
...
+  let bundleFiles: FileInfo[] = [];
   const cache = new Cache(CACHE_KEY, baseDir);
 
   // Filter for supported extensions/files only
   let processingFiles: string[] = filterSupportedFiles(files, supportedFiles);
 
   // Exclude files to be ignored based on ignore rules. We assume here, that ignore rules have not been changed.
-  processingFiles = processingFiles.filter(f => !isMatch(f, fileIgnores));
+  processingFiles = processingFiles
+    .map(f => resolveBundleFilePath(baseDir, f))
```
pkey, line 301: question: why is this extra step needed? Last time a change in the way we ignore things was made, we introduced  performance issues on the client.

src/files.ts
```diff
@@ -207,55 +207,72 @@ async function* searchFiles(
...
+      filterSupportedFiles(files, options.supportedFiles),
+      options.baseDir,
+      symlinksEnabled,
+      options.fileIgnores,
+    );
     for await (const filePath of searcher) {
-      const fileInfo = await getFileInfo(filePath.toString(), baseDir, false, cache);
-      if (fileInfo && fileInfo.size <= maxFileSize) {
+      const fileInfo = await getFileInfo(filePath.toString(), options.baseDir, false, cache);
+      // dc ignore AttrAccessOnNull: false positive, there is a precondition with &&
```
pkey, line 274: question: what does this mean?

src/bundles.ts
```diff
@@ -148,37 +158,103 @@ async function fullfillRemoteBundle(
...
+  fileIgnores: string[];
+}
+
+/**
+ * Creates a remote bundle and returns response from the bundle API
+ *
+ * @param {CreateBundleFromFoldersOptions} options
+ * @returns {Promise<FileBundle | null>}
+ */
+export async function createBundleFromFolders(options: CreateBundleFromFoldersOptions): Promise<FileBundle | null> {
```
pkey, line 218: required: I guess [this one](https://github.com/snyk/code-client/blob/master/src/analysis.ts#L441) can be deleted then?

tests/auth.api.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const response = await checkSession({
+      authHost,
+      draftToken: 'dummy-token',
+    });
+    expect(response.type).toEqual('success');
+    if (response.type === 'error') return;
+    expect(response.value).toEqual('');
+  });
+
+  // TODO: find a way to test successfull workflow automatically
```
pkey, line 34: required: let's put it to jira or stepsize as tech debt.

package.json
Arvi3d: default one was too small for some test repos. Since we don't want to have flaky tests, and we don't test performance here, i wanted to give big enough timeout

src/files.ts
Arvi3d: without this comment, our engine falsely reports about possible null value in expression "fileInfo.size"
It can't handle yet the fact, that there is a pre condition:  "fileInfo &&"

src/bundles.ts
Arvi3d: but it was delete.

tests/auth.api.spec.ts
Arvi3d: stepsize: Find a way to test successful Snyk authentication flow automatically

src/bundles.ts
pkey: My bad then, must have missed then in github diff. All good then!

tests/auth.api.spec.ts
Arvi3d: [stepsize] Find a way to test successful Snyk authentication flow automatically

src/interfaces/analysis-result.interface.ts
Arvi3d: done

src/files.ts
shinin-light: This should not bring performance degradation. However, the reason for the change is that in the previous api we were sending absolute paths in the bundles, while now we use relative paths instead (the consistency of the ignores should be guaranteed by the passing tests).

pkey (approved):

shinin-light:
`question`: are there any concerns about releasing this to master/npm while CLI and IDEs are not aligned yet? Should we double check that we are updating a major version and that their acceptance range is not including it?

Arvi3d:
> `question`: are there any concerns about releasing this to master/npm while CLI and IDEs are not aligned yet? Should we double check that we are updating a major version and that their acceptance range is not including it?

Yep, based on these:
- https://github.com/snyk/vscode-extension/blob/master/package.json#L278 
- https://github.com/snyk/snyk/blob/master/package.json#L79
there should be no issue

I'm going to upgrade and test code-agent right after, as there will be an issue otherwise:
https://github.com/snyk/code-agent/blob/master/package.json

PR #99, opened 2021-08-18
feat: exclude files from bundle with .snyk

saark-snyk:
- [x] Ready for review
- [x] Linked to Jira ticket - https://snyksec.atlassian.net/browse/PUL-157

#### What does this PR do?
Adds the ability to exclude files from the analysis bundle when adding them under the `exclude` section in `.snyk` file

Nothing specific 

Adding entries to a repo's `.snyk` file under the `exclude` field (same as in the sample-repo inside under `tests` folder) and making sure they're not sent for analysis.

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -466,8 +466,8 @@ export function parseDotSnykExcludes(pathToDotSnykFile: string): string[] {
   try {
     const dotSnykFile = fs.readFileSync(pathToDotSnykFile, 'utf-8');
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    const parsed: { exclude: string[] } = parseYaml(dotSnykFile);
-    return parsed.exclude.map(path => `${nodePath.dirname(pathToDotSnykFile)}/${path}`);
+    const parsed: { exclude: { code: string[] } } = parseYaml(dotSnykFile);
+    return parsed.exclude.code.map(path => `${nodePath.dirname(pathToDotSnykFile)}/${path}`);
   } catch (err) {
     // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
     if (err.code === 'EACCES' || err.code === 'EPERM') {
--- a/src/files.ts
+++ b/src/files.ts
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import fg from '@snyk/fast-glob';
 import multimatch from 'multimatch';
 import crypto from 'crypto';
+import { parse as parseYaml } from 'yaml';
 import union from 'lodash.union';
 import util from 'util';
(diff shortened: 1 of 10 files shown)


src/files.ts
```diff
@@ -248,29 +254,24 @@ export async function* collectBundleFiles({
...
   for (const folder of dirs) {
-    const searcher = searchFiles(globPatterns, folder, symlinksEnabled, options.fileIgnores);
+    const searcher = searchFiles(globPatterns, folder, symlinksEnabled, fileIgnores);
     // eslint-disable-next-line no-await-in-loop
     for await (const filePath of searcher) {
-      const fileInfo = await getFileInfo(filePath.toString(), options.baseDir, false, cache);
+      const fileInfo = await getFileInfo(filePath.toString(), baseDir, false, cache);
       // dc ignore AttrAccessOnNull: false positive, there is a precondition with &&
-      if (fileInfo && fileInfo.size <= maxPayload) {
+      if (fileInfo && fileInfo.size <= maxPayload && !fileExcludes.includes(fileInfo.bundlePath)) {
```
shinin-light, line 264: This would not work sadly. We've already tried to post-filter files based on exclusion rules, but it's too slow for too many customers. We need to inject the exclusion rules into the glob search in order to prune entire subtrees a-priori and not waste time scanning files one-by-one if we know they would be ignored anyway.
The whole distinction between `fileIgnores` and `fileExcludes` is irrelevant. All ignores and excludes at this level are in fact path exclusions, so please merge the two logics together (you will probably need to translate yaml exclusions to glob patterns).

shinin-light (changes requested):

src/files.ts
```diff
@@ -462,3 +461,22 @@ export function* composeFilePayloads(files: FileInfo[], bucketSize = MAX_PAYLOAD
...
+    const parsed: { exclude: string[] } = parseYaml(dotSnykFile);
+    return parsed.exclude.map(path => `${nodePath.dirname(pathToDotSnykFile)}/${path}`);
+  } catch (err) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if (err.code === 'EACCES' || err.code === 'EPERM') {
+      console.error(`${pathToDotSnykFile} is not accessible.`);
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if (err.code === 'ENOENT') {
+      console.error(`no such file or directory: ${pathToDotSnykFile}`);
```
shinin-light, line 478: How is this an error? Many directories won't have a `.snyk` file, at most it's an info level log.

shinin-light (approved):
Minor comment, but it looks good

PR #101, opened 2021-08-31
[NEB-124] fix: fixed request size for bundle requests

shinin-light:
[NEB-124]
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
- changed batching logic for bundle creation/extension/upload to take into consideration the real size of the expected request payload (file path + file hash + optional file content)
- added maximum parallelization limit on batch upload (using p-map) to ease network congestions
- added retry mechanism on 503 responses to avoid breaking the whole (possibly very long) extension flow because of a single network error
- tuned the tests accordingly

#### Additional questions

[NEB-124]: https://snyksec.atlassian.net/browse/NEB-124


src/bundles.ts
```diff
@@ -97,12 +101,13 @@ export async function uploadRemoteBundle({
   const apiParams = pick(options, ['baseURL', 'sessionToken', 'source', 'bundleHash']);
 
   const uploadFileChunks = async (bucketFiles: FileInfo[]): Promise<boolean> => {
-    const resp = await extendBundle({
+    let resp = await extendBundle({
```
ArturSnyk, line 104: why are we changing this to `let`? I don't see re-assigning of the object 🤔

shinin-light: just a typo. I needed to change it during some tests and then I forgot to revert it. Fixing it now

neonnoon (approved):
Looks good! 👍 I'm not quite certain about the retry functionality details.

src/bundles.ts
```diff
@@ -46,10 +51,9 @@ async function* prepareRemoteBundle({
 }: PrepareRemoteBundleOptions): AsyncGenerator<Result<RemoteBundle, BundleErrorCodes>> {
   let response: Result<RemoteBundle, BundleErrorCodes>;
   let { bundleHash } = options;
-
-  const fileChunks = chunk(options.files, maxPayload / 300);
-  emitter.createBundleProgress(0, fileChunks.length);
-  for (const [i, chunkedFiles] of fileChunks.entries()) {
+  let cumulativeProgress = 0;
+  emitter.createBundleProgress(cumulativeProgress, options.files.length);
+  for (const chunkedFiles of composeFilePayloads(options.files, maxPayload)) {
```
saark-snyk, line 56: Have we considered using a different loop than `for...of`? If we're talking optimizations, this is the slowest one...

src/constants.ts
```diff
@@ -11,6 +11,9 @@ export const DOTSNYK_FILENAME = '.snyk';
 export const EXCLUDED_NAMES = [GIT_FILENAME, GITIGNORE_FILENAME, DCIGNORE_FILENAME];
 export const CACHE_KEY = '.dccache';
 export const MAX_UPLOAD_ATTEMPTS = 5;
+export const UPLOAD_CONCURRENCY = 5;
+export const MAX_RETRY_ATTEMPTS = 5; // Request retries on 503 errors
+export const REQUEST_RETRY_DELAY = 30 * 1000; // 30 seconds delay between retries
```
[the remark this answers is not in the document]
shinin-light: I've modelled this on the actual response we get from our backend when performing such requests:
```
<= Response: 502 "\n<html><head>\n<meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\">\n<title>502 Server Error</title>\n</head>\n<body text=#000000 bgcolor=#ffffff>\n<h1>Error: Server Error</h1>\n<h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>\n<h2></h2>\n</body></html>\n"
```
So yes, it would take a max of 5 * 30s now. I'm up for suggestions on this ;)

src/needle.ts
```diff
@@ -52,26 +60,47 @@ export function makeRequest(payload: Payload): Promise<{ success: boolean; respo
...
       emitter.apiRequestLog(`<= Response: ${response.statusCode} ${JSON.stringify(response.body)}`.slice(0, 399));
-      const success = !!(response.statusCode && response.statusCode >= 200 && response.statusCode < 300);
-      return { success, response };
-    })
-    .catch(err => {
+      success = !!(response.statusCode && response.statusCode >= 200 && response.statusCode < 300);
+      if (success) return { success, response };
+      
+      // Try to avoid breaking requests due to temporary network errors
+      if (attempts > 1 && response.statusCode && [
```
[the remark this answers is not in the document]
shinin-light: I've just quickly read the errorCodes we are using in the constants and listed the ones that are not application-level errors.

src/bundles.ts
shinin-light: Do you have a specific suggestion?

src/bundles.ts
```diff
@@ -112,15 +118,17 @@ export async function uploadRemoteBundle({
...
-    return (await Promise.all(tasks)).some(r => !!r);
-  }
-  return true;
+  const results = await pMap(
+    tasks,
+    async (task: FileInfo[]) => await uploadFileChunks(task),
+    { concurrency: UPLOAD_CONCURRENCY }
+  );
+  // Returning false if at least one result is false
+  return !results.some(r => !r);
```
saark-snyk, line 131: `nit`: it's kind of my own preference but in order to avoid double negation you can use:
```suggestion
  return results.every(r => !!r);
```

src/needle.ts
```diff
@@ -52,26 +60,47 @@ export function makeRequest(payload: Payload): Promise<{ success: boolean; respo
...
-    open_timeout: 0,
     agent,
   };
 
   emitter.apiRequestLog(`=> HTTP ${method?.toUpperCase()} ${url} ${data ?? ''}`.slice(0, 399));
-
-  return needle(method, url, data, options)
-    .then(response => {
+  let attempts = payload.attempts || MAX_RETRY_ATTEMPTS;
+  let retryDelay = payload.retryDelay || REQUEST_RETRY_DELAY;
```
saark-snyk, line 76: `question`: since all of these are numbers, are you using `||` in order to take that into consideration? 
If not, I might suggest using `??` to avoid counting `0` as a valid value

src/bundles.ts
shinin-light: Like the idea!

src/bundles.ts
saark-snyk: You take a look at this:
 https://link.medium.com/Zr7kdVR0djb

tl;dr
If possible, the plain old `for` would be the fastest

PR #106, opened 2021-09-16
fix: dot-snyk files respected

Spoor2709:
- [ ] Ready for review

#### What does this PR do?

This PR now allows the everything using code-client to use .snyk file

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -7,7 +7,15 @@ import { parse as parseYaml } from 'yaml';
 import union from 'lodash.union';
 import util from 'util';
 import { Cache } from './cache';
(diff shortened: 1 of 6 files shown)


src/files.ts
```diff
@@ -98,7 +98,13 @@ export function parseFileIgnores(path: string): string[] {
   const dirname = nodePath.dirname(path);
   try {
     const f = fs.readFileSync(path, { encoding: 'utf8' });
+    if (path.includes('snyk')) {
```
novalex, line 101: Should be `DOTSNYK_FILENAME` since we have a constant for it.

novalex (approved):

PR #108, opened 2021-09-24
Feat/legacy json support

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

This PR is a pre-requisite to update VSC to use new Code API, instead of the old one.
analysis option now accepts a new key legacy?: boolean to control results' format.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -12,6 +12,7 @@ export const EXCLUDED_NAMES = [GIT_FILENAME, GITIGNORE_FILENAME, DCIGNORE_FILENA
 export const CACHE_KEY = '.dccache';
 export const MAX_UPLOAD_ATTEMPTS = 5;
 export const UPLOAD_CONCURRENCY = 5;
+export const POLLING_INTERVAL = 500;
 export const MAX_RETRY_ATTEMPTS = 5; // Request retries on network errors
 export const REQUEST_RETRY_DELAY = 30 * 1000; // 30 seconds delay between retries
 
--- a/src/needle.ts
+++ b/src/needle.ts
@@ -3,7 +3,7 @@ import needle from 'needle';
 import * as querystring from 'querystring';
 import https from 'https';
 import { URL } from 'url';
-import emitter from './emitter';
+import { emitter } from './emitter';
 
 import {
   ErrorCodes,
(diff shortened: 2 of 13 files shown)


src/constants.ts
```diff
@@ -12,6 +12,7 @@ export const EXCLUDED_NAMES = [GIT_FILENAME, GITIGNORE_FILENAME, DCIGNORE_FILENA
 export const CACHE_KEY = '.dccache';
 export const MAX_UPLOAD_ATTEMPTS = 5;
 export const UPLOAD_CONCURRENCY = 5;
+export const POLLING_INTERVAL = 1000;
```
pkey, line 15: question: looks like polling interval was changed from 500 to 1000. Any reason for that?

tests/analysis.spec.ts
```diff
@@ -152,6 +155,28 @@ describe('Functional test of analysis', () => {
...
+        fileOptions: {
+          paths: [sampleProjectPath],
+          symlinksEnabled: false,
+          maxPayload: 1000,
+          defaultFileIgnores: undefined,
+        },
+      });
+
+      expect(bundle).toBeTruthy();
+      if (!bundle) return; // TS trick
```
pkey, line 171: suggestion: shall we just use @ts-ignore or sth along those line instead?

pkey (approved):
Couple of qs, otherwise lgtm

src/constants.ts
Arvi3d: ah, probably i did it by mistake, I thought it was 1 sec.
There was not much thinking on choosing it in the first place too.
I guess i can revert it back to 500ms

tests/analysis.spec.ts
Arvi3d: not sure i get you. can you write a suggestion code here?

src/constants.ts
Arvi3d: done

src/needle.ts
```diff
@@ -78,10 +78,10 @@ export async function makeRequest(payload: Payload): Promise<{ success: boolean;
   do {
     try {
       response = await needle(method, url, data, options);
-      emitter.apiRequestLog(`<= Response: ${response.statusCode} ${JSON.stringify(response.body)}`.slice(0, 399));
+      emitter.apiRequestLog(`<= Response: ${response.statusCode} ${JSON.stringify(response.body)}`);
```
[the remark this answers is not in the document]
michelkaporin: Seeing this in the console since update:
Screenshot 2021-10-08 at 08 57 33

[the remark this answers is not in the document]
Arvi3d: ah, i probably forgot to limit it inside vsc. 
The idea was that we need to be able to debug this data.
And i was going to do this limiting inside vsc.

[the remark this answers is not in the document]
michelkaporin: @Arvi3d got it, thanks. I'll do it on VSC side myself 👍

PR #109, opened 2021-09-26
[PUL-235]fix: excludes with code or gloabl

Spoor2709:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?
This pr fixes the bug that you need to have a global and code field under excludes in the .snyk file to work

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -108,10 +108,12 @@ export function parseFileIgnores(path: string): string[] {
     const f = fs.readFileSync(path, { encoding: 'utf8' });
     if (path.includes(DOTSNYK_FILENAME)) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const parsed: { exclude: { code: string[]; global: string[] } } = parseYaml(f);
+      const parsed: { exclude: { code?: string[]; global?: string[] } } = parseYaml(f);
 
       const concatIgnorePath = (p: string) => `${nodePath.dirname(path)}/${p}`;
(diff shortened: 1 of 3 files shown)


orkamara (changes requested):
Change looks great! can we add a test? 🙏

src/files.ts
```diff
@@ -108,10 +108,12 @@ export function parseFileIgnores(path: string): string[] {
     const f = fs.readFileSync(path, { encoding: 'utf8' });
     if (path.includes(DOTSNYK_FILENAME)) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const parsed: { exclude: { code: string[]; global: string[] } } = parseYaml(f);
+      const parsed: { exclude: { code?: string[]; global?: string[] } } = parseYaml(f);
 
       const concatIgnorePath = (p: string) => `${nodePath.dirname(path)}/${p}`;
-      return [...parsed.exclude.code.map(concatIgnorePath), ...parsed.exclude.global.map(concatIgnorePath)];
+      const codeIgnoredPaths = parsed.exclude.code?.map(concatIgnorePath) || [];
```
saark-snyk, line 114: `nit`: Just a personal thing, but I usually use `??` instead of `||`, and use `||` only when I need to take into account `0` in my condition. But, that's my own personal thing so no biggy :)

saark-snyk (approved):

orkamara (approved):

PR #110, opened 2021-10-12
chore: optimization of file uploading of big projects

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?
bunch of improvements to make it more predictable and performant while uploading big projects

- removed an argument "maxPayload" from public function, as it's an internal details of our server details
- optimized several places with exceeding memory use with "reduce" function
- adjusted retry attempts counters
- optimized file uploading by creating chunked bundles
- fixed .snyk parsing to be able to process folders without trailing "**"

This PR depends on another PR and will have broken CircleCI test until that one is merged:

https://github.com/snyk/octopus/pull/1108


tests/sample-repo/.snyk
```diff
@@ -1,5 +1,5 @@
 exclude:
   global:
-    - exclude/excluded-folder/**
```
novalex, line 3: IMO we should also have a test case for paths without the leading slash - i.e. between excluding just the root `exclude/excluded-folder` and excluding any nested `exclude/excluded-folder`.

src/bundles.ts
```diff
@@ -87,46 +86,40 @@ interface UpdateRemoteBundleOptions extends ConnectionOptions {
...
       ...apiParams,
-      files: bucketFiles.reduce((d, f) => ({ ...d, [f.bundlePath]: pick(f, ['hash', 'content']) }), {}),
+      files: bucketFiles.reduce((d, f) => {
+        d[f.bundlePath] = pick(f, ['hash', 'content']);
+        return d;
+      }, {}),
     });
 
-    // During upload process, we expect the bundleHash not to change (same file map)
-    if (resp.type !== 'error' && resp.value.bundleHash === apiParams.bundleHash) {
```
shinin-light, line 106: I don't agree with this change (and all of its consequences, like the removal of the tests and the final bundle check). I added this check specifically because we had cases where the upload request was creating a new bundle and we don't take the returning `resp.value.bundleHash` into consideration in future steps. We currently use the ExtendBundle API both for extending the bundle definition (list of files) and to upload files once the definition has been extended. Since we want this distinction (to take advantage of the `missingFiles`) but we are using the same API to achieve both, we need to consider the possibility that at the upload step we are actually creating a new bundle instead of just extending it into itself (i.e. when the returned `bundleHash` is the same as the requested `bundleHash`).

src/bundles.ts
```diff
@@ -139,14 +132,11 @@ async function fullfillRemoteBundle(options: FullfillRemoteBundleOptions): Promi
...
   while (remoteBundle.missingFiles.length && attempts < (options.maxAttempts || MAX_UPLOAD_ATTEMPTS)) {
     const missingFiles = await resolveBundleFiles(options.baseDir, remoteBundle.missingFiles);
-    const isUploaded = await uploadRemoteBundle({
+    await uploadRemoteBundle({
       ...connectionOptions,
       bundleHash: remoteBundle.bundleHash,
       files: missingFiles,
     });
-    if (!isUploaded) {
-      throw new Error('Failed to upload some files');
```
shinin-light, line 148: Same as above

shinin-light (changes requested):

src/bundles.ts
Arvi3d: "removal of the tests" what do you mean here? i did not remove tests. maybe it was some specific commit, where i commented tests just temporarly

PR #112, opened 2021-10-15
fix: resolve relative bundle file path correctly [ROAD-329]

michelkaporin:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
This PR resolves the issue of determining relative bundle file paths when many folders are passed for the analysis.

When `getBundleFilePath()` is called for multiple folders analysis (e.g. VS Code workspace with many folders), it is passed with `baseDir = ''` argument that makes relative path resolution fail on Linux within VS Code extension currently.

The following behaviour is observed at the moment:
args: `baseDir='', filePath='/home/user/Downloads/goof-master/routes/index.js'`
returns: `'Downloads/goof-master/routes/index.js'`

`path.relative()` relies on `process.cwd()` that can be different based on where Node.js process was started. For VS Code, If this code is called on Mac OS, `process.cwd()` defaults to `/`, however Linux defaults to the user directory e.g. `/home/user`. Thus wrong analysis results file paths for bundles will be returned. 

Relevant Jira ticket: https://snyksec.atlassian.net/browse/ROAD-329

### DIFF
--- a/tests/files.spec.ts
+++ b/tests/files.spec.ts
@@ -8,6 +8,7 @@ import {
   composeFilePayloads,
   parseFileIgnores,
   getFileInfo,
+  getBundleFilePath,
 } from '../src/files';
 
 import { sampleProjectPath, supportedFiles, bundleFiles, bundleFilesFull, bundleFileIgnores } from './constants/sample';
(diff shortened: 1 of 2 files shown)


tests/files.spec.ts
```diff
@@ -113,4 +114,16 @@ describe('files', () => {
...
+
+  it('obtains correct bundle file path if no baseDir specified', () => {
+    const baseDir = '';
+    const darwinPath = '/Users/user/Git/goof/routes/index.js';
+    expect(getBundleFilePath(darwinPath, baseDir)).toEqual(darwinPath);
+
+    const linuxPath = '/home/user/Git/goof/routes/index.js';
+    expect(getBundleFilePath(linuxPath, baseDir)).toEqual(linuxPath);
+
+    const windowsPath = 'C:\\Users\\user\\Git\\goof\\index.js';
```
Arvi3d, line 126: i'm afraid our engine will fail, if it receives this path.
can you please also add an explicit test for it?

Arvi3d: ah, stop, this is a reverse process. all good

Arvi3d (approved):

PR #116, opened 2021-11-24
fix: considering customers lang support NEBULA-174

metju90:
#### What does this PR do?

Certain customers do not have all code languages enabled. Currently, the code-client only checks supported languages from (DC) system. This is not enough, because if for an example, C and C++ are not enabled for a given customer and those 2 languages are supported by DC system, we would scan files for the mentioned languages. This PR solves this issue.

The above issue is reproducible by the customer of code-client -- Code Agent

Jira: https://snyksec.atlassian.net/jira/software/c/projects/NEBULA/boards/282?modal=detail&selectedIssue=NEBULA-174

### DIFF
--- a/src/bundles.ts
+++ b/src/bundles.ts
@@ -2,7 +2,7 @@
 import pick from 'lodash.pick';
 import omit from 'lodash.omit';
 import pMap from 'p-map';
-
+import langMap from 'lang-map';
 import { BundleFiles, FileInfo, SupportedFiles } from './interfaces/files.interface';
 
 import {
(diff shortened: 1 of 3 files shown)


src/bundles.ts
```diff
@@ -249,3 +261,32 @@ export async function createBundleFromFolders(options: CreateBundleFromFoldersOp
...
+
+function getAllowedExtensions(systemExtensions: string[], customerExtensions: string[]): string[] {
+  const allowedExtensions = customerExtensions.filter(ce => {
+    if (systemExtensions.includes(ce)) {
+      return true;
+    }
+    return false;
+  });
+  return allowedExtensions;
+}
```
metju90, line 292: @Arvi3d Bundle's supported file extension is a list of 30 items.

This means that if we are using a third-party library, we must make sure that it has the same (or more, but not less) numbers of extensions that Bundle supports. Otherwise, certain scans might not happen because `getAllowedExtensions()` will exclude their file extension. 

Full list of extensions provided by Bundle through code-client:
```
[
      '.aspx',
      '.c',
      '.cc',
      '.cls',
      '.cpp',
      '.cs',
      '.cxx',
      '.ejs',
      '.erb',
      '.es',
      '.es6',
      '.go',
      '.h',
      '.haml',
      '.hpp',
      '.htm',
      '.html',
      '.hxx',
      '.java',
      '.js',
      '.jsx',
      '.php',
      '.phtml',
      '.py',
      '.rb',
      '.rhtml',
      '.slim',
      '.ts',
      '.tsx',
      '.vue',
    ];
   ```

PR #114, opened 2021-10-28
[COD-571] fix: refactored API calls to make them tolerate network failures and retry

Arvi3d:
https://snyksec.atlassian.net/browse/COD-571

- [ ] Ready for review
- [ ] Linked to Jira ticket

PR #119, opened 2021-12-23
fix: getIpFamily not detecting IPv6 support

michelkaporin:
- [x] Ready for review
- [x] Linked to Jira ticket - https://snyksec.atlassian.net/browse/ROAD-604

#### What does this PR do?
Fixes regression that got introduced as part of [this PR](https://github.com/snyk/code-client/pull/114/files), where try/catch [was removed](https://github.com/snyk/code-client/pull/114/files#diff-d12299d8c4ec4ce4401d0a4ca90f330520ad90714c71841b145d069010e06e5cL117) within `getIpFamily` function.

This led to all users that support IPv6 default to IPv4 protocol when authenticating through IDE, since the `verify/callback` always returns 401 at this point (i.e. `res.success = false`).

Since the new code doesn't bubble errors up to the caller, added additional map for the exception that gets thrown when IPv6 lookup is not supported by a user.

https://github.com/snyk/vscode-extension/issues/116


src/constants.ts
```diff
@@ -48,6 +48,7 @@ export const NETWORK_ERRORS = {
   ECONNRESET: ErrorCodes.connectionRefused,
   ENETUNREACH: ErrorCodes.connectionRefused,
   ENOTFOUND: ErrorCodes.dnsNotFound,
+  EADDRNOTAVAIL: ErrorCodes.dnsNotFound, // happens when DNS cannot resolve the IPv6
```
ArtsiomCh, line 51: unused?

src/http.ts
```diff
@@ -99,7 +99,10 @@ export async function getIpFamily(authHost: string): Promise<IpFamily> {
     },
     0,
   );
-  return res.success ? family : undefined;
+
+  const ipv6Incompatible = (<FailedResponse>res).errorCode == ErrorCodes.dnsNotFound;
```
ArtsiomCh, line 103: could we have that `ErrorCodes.dnsNotFound` for any other reason rathe then piv6 compatibility? if so, is `undefined` return will correct one for such cases?

ArtsiomCh:
failed tests in CI? `default_workflow`

src/constants.ts
michelkaporin: @ArtsiomCh this gets thrown from the needle request if IPv6 cannot be used.

src/http.ts
michelkaporin: Yes. If ENOTFOUND is thrown, this will be true but will be also an indication that it's incompatible with IPv6.

Since `ErrorCodes.dnsNotFound` is a purely a customised error code that gets assigned only within try/catch, the handling will be the consistent with the current CLI IPv6 incompatibility handling.

michelkaporin:
> failed tests in CI? default_workflow

haven't changed anything around the code that fails, thus I assume it's the same issue that happens from time to time due to backend Snyk Code changes. It's out of scope of this PR to fix, unless Sast team asks me to 😄

src/constants.ts
ArtsiomCh: May be it my lack of TS knowledge, but I can't see usage of `EADDRNOTAVAIL` anywhere in the code...

ArtsiomCh (approved):
LGTM, but I didn't run and/or test it.
Ooch, just found it's not vsc plugin repo... Then my approval could be just ignored.

saark-snyk (approved):

src/http.ts
saark-snyk, line 103: why the shallow comparison? (`==`)

michelkaporin: Knew that these two are always of number type. Changed it to strict equality to be future-proof here ✅

PR #120, opened 2022-01-14
fix: getIpFamily - do not expect specific error for clients without IPv6

michelkaporin:
- [x] Ready for review
- [x] Linked to Jira ticket - https://snyksec.atlassian.net/browse/ROAD-643

#### What does this PR do?
This PR is a replacement for [previous PR](https://github.com/snyk/code-client/pull/119) to fix the regression for IPv6 clients. We've noticed that some of the clients who don't have full IPv6 supports started to authenticate using IPv6 which broke their VS Code extension.

Some users encountered ENETUNREACH error instead of EADDRNOTAVAIL as wrongly assumed in the previous fix PR. Thus we shouldn't rely on specific error, rather on an error. This is how it was before the refactoring around http took place and is consistent with Snyk CLI code.

This fixes the following issues:
https://snyk.slack.com/archives/C026PS65XA8/p1642092098003500
https://snyk.slack.com/archives/C01FA88CUU9/p1641979366003300?thread_ts=1641978929.003100&cid=C01FA88CUU9
https://github.com/snyk/vscode-extension/issues/132 🤞

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -48,7 +48,6 @@ export const NETWORK_ERRORS = {
   ECONNRESET: ErrorCodes.connectionRefused,
   ENETUNREACH: ErrorCodes.connectionRefused,
   ENOTFOUND: ErrorCodes.dnsNotFound,
-  EADDRNOTAVAIL: ErrorCodes.dnsNotFound, // happens when DNS cannot resolve the IPv6
 };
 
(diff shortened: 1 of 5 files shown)


saark-snyk (commented):
would appreciate adding some tests to verify this behavior

saark-snyk (approved):

PR #123, opened 2022-02-03
fix: glob matching in extending bundle on Windows [ROAD-650]

michelkaporin:
- [x] Ready for review
- [x] Linked to Jira ticket - https://snyksec.atlassian.net/browse/ROAD-650

#### What does this PR do?
Glob determines changed files as removed, when bundle gets extended currently on Windows environments.

The problem arises due to a use of backslashes when passing glob pattern to fast-glob library. As [fast-glob documentation states](https://github.com/snyk/fast-glob#how-to-write-patterns-on-windows), forward slashes should be used instead on Win environments.

It would be great to run automates tests on Windows environment to pick up those type of issues in future. I expect "extend bundle files" test should pick the problem up, when tested under Windows.

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -302,6 +302,10 @@ export async function prepareExtendingBundle(
 
   if (processingFiles.length) {
     // Determine existing files (minus removed)
+    if (isWindows) {
+      processingFiles = processingFiles.map(f => f.replace(/\\/g, '/')); // fg requires forward-slashes in Windows globs
+    }
+
     const entries = await fg(processingFiles, {
       ...fgOptions,
       cwd: baseDir,


src/files.ts
```diff
@@ -302,6 +302,10 @@ export async function prepareExtendingBundle(
 
   if (processingFiles.length) {
     // Determine existing files (minus removed)
+    if (isWindows) {
+      processingFiles = processingFiles.map(f => f.replace(/\\/g, '/')); // fg requires forward-slashes in Windows globs
```
pkey, line 306: [stepsize] we don't have test coverage for window specific behaviour. rare case but possible.

src/files.ts
```diff
@@ -302,6 +302,10 @@ export async function prepareExtendingBundle(
 
   if (processingFiles.length) {
     // Determine existing files (minus removed)
+    if (isWindows) {
```
pkey, line 305: suggestion: @michelkaporin do you think it would make sense to write a test case for this (simulating the `isWindows` paramteter)?

michelkaporin: It's difficult since current tests are dependent on environment when working with bundle from what I see.
As mentioned in the PR description I think we should simply have a CircleCI agent that runs our tests under Windows environment. 

In fact [this test](https://github.com/snyk/code-client/blob/a4ffee2193889ddf8e408a247b66cfe4c37df53a/tests/files.spec.ts#L57) should have picked up the problem in my understanding, if the test would be run on Windows machine.

pkey: @michelkaporin hmm, that makes sense. Ok, we might explore it later but won't be blocking you over this.

pkey (approved):

PR #124, opened 2022-02-03
test: remove precise length for results

pkey:
Here we don't really care about the length of results and having a set
number for amount of issues makes this test prone to flakiness.  We are more
interested whether the bundle was extended properly.

### DIFF
--- a/tests/analysis.spec.ts
+++ b/tests/analysis.spec.ts
@@ -163,8 +163,8 @@ describe('Functional test of analysis', () => {
       expect(bundle.analysisResults.type === 'legacy').toBeTruthy();
       if (bundle.analysisResults.type !== 'legacy') return;
 
-      expect(Object.keys(bundle.analysisResults.files)).toHaveLength(5);
-      expect(Object.keys(bundle.analysisResults.suggestions)).toHaveLength(11);
+      expect(Object.keys(bundle.analysisResults.files).length).toBeGreaterThan(0);
+      expect(Object.keys(bundle.analysisResults.suggestions).length).toBeGreaterThan(0);
     });
 
     it('analyze folder - with sarif returned', async () => {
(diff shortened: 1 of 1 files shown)


neonnoon:
To be honest, I'm not sure I understand what those tests are doing. Are we testing the content that the API returns (i.e., we're testing for the contract), or are we testing local functionality?

michelkaporin (approved):
Since this is blocking the release currently and I had to change the tests expectations back and forth several times in my last PRs -> LGTM.

neonnoon (approved):
Approving so we're not blocking, but I think we need to improve our testing principles here.

pkey:
> Why is the test flaky, does it run against codeapi rather than a mock endpoint?

Yes, it does run against `code_api`. To best of my understanding, this is testing whether the files were extended once the endpoint was called... But it means we involve a third party service, which is prone to flakiness.

pkey:
> To be honest, I'm not sure I understand what those tests are doing. Are we testing the content that the API returns (i.e., we're testing for the contract), or are we testing local functionality?

@neonnoon It uploads the files, then, using a thing called extender ([here](https://github.com/snyk/code-client/blob/master/tests/constants/sample.ts#L55)), manipulates the content of files in memory and then tests whether newly uploaded files are being handled correctly by the API.

So, I'd say it does both - tests the contract and tests local functionality. Probably should focus on the latter in this case.

pkey:
That being said, I don't think any of these tests really care about the amount of issues found - I assume we trust what, given the files provided, `suggest` will return.

neonnoon:
[stepsize] Can we separate api contract testing from local unit tests?

tests/analysis.spec.ts
```diff
@@ -163,8 +163,8 @@ describe('Functional test of analysis', () => {
       expect(bundle.analysisResults.type === 'legacy').toBeTruthy();
       if (bundle.analysisResults.type !== 'legacy') return;
```
neonnoon, line 165: [stepsize] Can we separate api contract testing from local unit tests?

pkey: Next step: bring stepsize back to our lives :D

neonnoon: Yes! 😬  One step at a time.

PR #127, opened 2022-02-16
fix: monoliths repo no longer timeouts

metju90:
Increasing connection timeout - one of the things needed for monoliths

### DIFF
--- a/src/needle.ts
+++ b/src/needle.ts
@@ -18,7 +18,7 @@ export declare interface Global extends NodeJS.Global {
 }
 declare const global: Global;
 
-const TIMEOUT_DEFAULT = 10000;
+const TIMEOUT_DEFAULT = 600000;
 
 const agentOptions = {
   keepAlive: true,


neonnoon (approved):

pkey:
@metju90 you mentioned that `requestId` helped you to discover this limit. Could you elaborate more how did you came to the conclusion here?

Also, do we know that at some point we won't encounter a repo which will surpass even that threshold and will be completely legit case? I guess, providing this is polling, we shouldn't put a limit at all and make sure that `api` always returns an error, when there is one.

metju90:
@pkey 
> you mentioned that `requestId` helped you to discover this limit.

That's not exactly what happened. Showing the actual logs (Pr merged 2 weeks ago in CLI) helped me understand that there is a connection timeout in upload process When I increased this limit and rescanned with monoliths, files were upload was successful, confirming the timeout was the problem.

What the `requestId` is helping me with is discovered the second problem with scanning monoliths (more details in Slack tomorrow)

> we shouldn't put a limit at all and make sure that `api` always returns an error
I think a request over the network can never be guaranteed to be delivered. If we remove the limit we can have calls hanging forever (since its possible that call hangs before it reaches `api`).

A safer thing to do is to know our limits and communicate it to customers

PR #140, opened 2022-03-29
feat: increase file limit to 1MB [NEBULA-456]

patricia-v:
#### What does this PR do?
Increases analysis file size limit to 1MB in order to reduce discrepancy between registry and CLI.

[Jira ticket](https://snyksec.atlassian.net/browse/NEBULA-456)


tests/api.spec.ts
```diff
@@ -245,8 +247,8 @@ describe('Requests to public API', () => {
       });
       expect(response.type).toEqual('success');
       if (response.type !== 'success') return; // TS trick
-      expect(response.value.bundleHash).toContain('77d2fc40d9b77f3e8cfc3b5046e634e5edd958ea3a9b00fb12051fa30942d61f');
-      expect(response.value.missingFiles).toHaveLength(11);
+      expect(response.value.bundleHash).toContain('06c8969ce7bce4c62e28f96cf9fc54a68cf25644aebd53be09094f365058f4a6');
+      expect(response.value.missingFiles).toHaveLength(12);
```
metju90, line 251: How did the missing files increase? we just replaced `main.js` and  with `big-file.js`, correct?

I've noticed this test was broken for a while, so if this is not making sense its not your fault 🙃

metju90 (approved):

tests/api.spec.ts
patricia-v: I think that since we're using the same bundle for most of these tests (with hash `fakeBundleHashFull`), when the test `extends bundle successfully` is done, it adds the new `big-file.js` to the missing files

patricia-v: The file `main.js` was not replaced, it was kept in the repo

PR #141, opened 2022-03-31
feat: track skipped files over max size [NEBULA-481]

patricia-v:
#### What does this PR do?
Adds logging in debug mode to CLI to warn when files are skipped due to their size and lists which ones were skipped.
This is the code-client side where the tracking of the skipped files was added.

Also, the file size of `big-file.js` in the test's sample repo was reduced by about 1MB since it didn't have to be as big. However, it is still over the max file size limit and will be skipped in the analysis.

[Link to CLI side](https://github.com/snyk/cli/pull/3083)

[NEBULA-481](https://snyksec.atlassian.net/browse/NEBULA-481)

#### Screenshots
image


src/bundles.ts
```diff
@@ -242,7 +244,7 @@ export async function createBundleFromFolders(options: CreateBundleFromFoldersOp
     supportedFiles,
   });
   for await (const f of bundleFileCollector) {
-    bundleFiles.push(f);
+    typeof f == 'string' ? skippedOversizedFiles.push(f) : bundleFiles.push(f);
```
metju90, line 247: So when `file` is a string, does it mean that it's a skipped file? 😮

metju90 (approved):
Nice job there, seems straightforward.

By the way, the *single* ternary operator saves you some lines of code at the cost of making it slightly harder to read 😄 . Nested ternary would be a no-go, but single ones are a bit more tolerable.

src/bundles.ts
patricia-v: Yes, since `collectBundleFiles` yields the `fileInfo` only for the files smaller than the max file size, and I've only added the string to return the path of the files that exceeded this max file size 😄

patricia-v:
> Nice job there, seems straightforward.
> 
> By the way, the _single_ ternary operator saves you some lines of code at the cost of making it slightly harder to read 😄 . Nested ternary would be a no-go, but single ones are a bit more tolerable.

I absolutely agree, I just hate to see if else's with one line each 😅

src/files.ts
```diff
@@ -247,8 +247,8 @@ export async function* collectBundleFiles({
     // Check if symlink and exclude if requested
     if (!fileStats || (fileStats.isSymbolicLink() && !symlinksEnabled)) continue;
 
-    if (fileStats.isFile() && fileStats.size <= MAX_FILE_SIZE) {
-      files.push(path);
+    if (fileStats.isFile()) {
+      fileStats.size <= MAX_FILE_SIZE ? files.push(path) : yield path;
```
pkey, line 251: question: you sure this doesn't break anything? Previously we would continue even if file would be bigger than the limit and now we are returning here.

src/bundles.ts
pkey: not sure I feel comfortable with this idea. May I suggest that instead of having this logic, we would check file `size` (it looks like it exists on FileInfo object) attribute here (on line `247`) and `push` based on that? It would be an extra check, yes, but much cleaner, in my opinion.

The only place where I see this is problematic, is [here](https://github.com/snyk/code-client/pull/141/files#diff-4a1b3df9d5079883d088444acd4ae285dfda7e14077739c1741c8f735b69a55eR251) because `fileStats` is a different object which doesn't correspond to FileInfo type. I have asked a question below regarding this.

patricia-v: > not sure I feel comfortable with this idea. May I suggest that instead of having this logic, we would check file size (it looks like it exists on FileInfo object) attribute here (on line 247) and push based on that? It would be an extra check, yes, but much cleaner, in my opinion.

So are you saying that we should delete the check for the max file size in `collectBundleFiles` and check it here?

> The only place where I see this is problematic, is [here](https://github.com/snyk/code-client/pull/141/files#diff-4a1b3df9d5079883d088444acd4ae285dfda7e14077739c1741c8f735b69a55eR251) because fileStats is a different object which doesn't correspond to FileInfo type. I have asked a question below regarding this.

This shouldn't necessarily be a problem here because `collectBundleFiles` will never yield `fileStats`, only `FileInfo` types (and now strings). I'll explain a bit more about this in your question below.

I agree that this is not the cleanest of the solutions and I really want it to be prettier, in my opinion the idea of checking the file size here wouldn't be too bad, it would only affect some of the tests.

src/files.ts
patricia-v: I don't think it would break since what happens after this is the sanitisation of these files, which basically gets the `FileInfo` from the files. This is necessary for the files that need to be scanned.
However, since I don't really need the `FileInfo` from the files that are bigger than the max file size, I can just return its path here and there's no need for sanitising.

So the only thing that this is adding is another 'pause' to the generator for the files that exceed the max size, that translates in having more values returned from the function in `bundles.ts` on the for cycle that goes through the collector, but it doesn't necessarily break anything.

src/bundles.ts
pkey: > So are you saying that we should delete the check for the max file size in collectBundleFiles and check it here?

I wouldn't change the existing logic; I was saying that for this specific use case we could check the size one more time. This doesn't seem like it would be an expensive operation but it would be cleaner than returning a string in my opinion.

Also, checking the size inside the function has a benefit of returning (or yielding in this case) value earlier.

src/files.ts
pkey: Thank you for explanation, I understand.

pkey (approved):
@patricia-v thanks for the answers, I think we can go with this solution. One thing I would do is to expand on the JSDoc of the function, explaining what the returned path (not the whole object) means, so that the caller of the function would easily understand what is going on.

src/bundles.ts
patricia-v: > I wouldn't change the existing logic; I was saying that for this specific use case we could check the size one more time.

But why check it here if no file greater than the max file is returned by the generator? 🤔 I need the list of files that were skipped due to their size, and I don't see how that could be achieved by leaving `collectBundleFiles` the way it was and also check for the size here, since all the files from `collectBundleFiles` will automatically be less than the max size... or am I misunderstanding something? 😅

pkey: you're right, yeah, that doesn't make sense. So if not for the first line which deals with `fileStats` object, it would be possible to keep the same interface, right?

patricia-v: Yes and no. I will try to simplify `collectBundleFiles` the best I can from my understanding:

1. From the paths provided, check through `fileStats` if it's a file or a dir
1.1. If the path is a file, checks if size is bigger than `MAX_FILE_SIZE` and, if not, adds it to `files`
1.2. If the path is a directory, adds it to `dirs`
2. Takes the directories in `dirs`, scans it and gets the `fileInfo` of each file, and checks if its size is bigger than `MAX_FILE_SIZE` - if not, provides the `fileInfo`
3. Takes the files in `files`, gets each file's `fileInfo`, and checks if its size is bigger than `MAX_FILE_SIZE` - if not, provides the `fileInfo`

One of the first things I've noticed is that when the path is a file, there's a double verification of its size in steps 1 and 3, which doesn't seem to make much sense, but I think it's just for redundancy.

However, if this check wasn't made in step 1, we could do what you're saying and return just the `fileInfo` of everything, but then we would have a couple of problems:

- We would be getting extra `fileInfo` that we wouldn't necessarily need because these files wouldn't be scanned in any way, I just need a string to list them in the debug mode. I don't know what impacts this could have in performance, if any.

- How would we differentiate the files that were to be scanned vs the ones that were too big? The `FileInfo` interface has the following fields: `filePath: string`, `bundlePath: string`, `size: number`, `hash: string` and `content?: string`. For the bigger files, we could create a mock version of a `fileInfo` that only contained the correct `filePath`/`bundlePath`, and then fill the `size` and `hash` with something that would tell us that it's a file too big, but I don't support this idea since we're taking things out of their scope. Or we could change the `FileInfo` interface and add an optional flag that would tell us if the file is too big to be scanned, but then why change this interface and not the return of the generator function?

pkey: Thanks for the clarification. Regarding your last point, which I think poses the main question, is exactly what I have been pondering - whether we should change the interface, or keep it and, as I suggested first, verify whether that file is oversized by double checking with `MAX_SIZE`.

As I said, I don't have a strong opinion here on why we shouldn't use `string` other than that it feels weird.

pkey: > But with this we go back to skipping the checks altogether in collectBundleFiles, right?

I mean, if we keep FileInfo, you said we won't know which files were scanned, which were oversized. We could then theoretically check the `size` property to figure that out. We don't need to skip the checks in `collectBundleFiles`, we could keep them but always return the FileInfo.

The more I talk about it the stranger it sounds but it would avoid returning a `string` 🤷🏻

patricia-v: The thing I'm a bit confused about your logic is that you're saying we'll do the checks in `collectBundleFiles` and in `bundles.ts`, but this would mean that in `collectBundleFiles` we would have something like this:

`fileInfo.size <= MAX_FILE_SIZE ? yield fileInfo : yield fileInfo;`

which is just ridiculous 😆 what I'm saying is that, at most, we would just do:

`if (fileInfo) { yield fileInfo }`

and then in `bundles.ts` something like this:

`f.size <= MAX_FILE_SIZE ? bundleFiles.push(f) : skippedOversizedFiles.push(f);`

metju90 (approved):

PR #142, opened 2022-04-01
fix: bundle file path resolution for paths with whitespace [ROAD-637]

michelkaporin:
- [ ] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
This PR addresses https://github.com/snyk/vscode-extension/issues/165 and https://github.com/snyk/vscode-extension/issues/114 respectively, where file navigation breaks when Snyk Code analysis is run in workspace (multi-folder) mode.

If multiple folders are passed for analysis, [`baseDir` gets set to empty string](https://github.com/snyk/code-client/blob/a8c8b58ab5da76663347f1ebdb3b86623459b6c2/src/files.ts#L177). It [skips result file paths normalisation](https://github.com/snyk/code-client/compare/fix/workspace-resolveBundleFilePath?expand=1#diff-e369c18a244b57d8ee4a1bdcabbea4cf17a6086c6f140efb494b125fa911c5a2L85), leading to encoded path provided back to the analysis issuer (VS Code extension in this case). VSCE doesn't do any file normalisation since relying on `code-client`, thus breaking file navigation for our end users.

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -83,14 +83,11 @@ export async function analyzeBundle(options: GetAnalysisOptions): Promise<Analys
 }
 
 function normalizeResultFiles(files: AnalysisFiles, baseDir: string): AnalysisFiles {
-  if (baseDir) {
-    return Object.entries(files).reduce((obj, [path, positions]) => {
-      const filePath = resolveBundleFilePath(baseDir, path);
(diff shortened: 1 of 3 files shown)


pkey (approved):
Looks good to me. I couldn't trace back why this check was introduced in the first place though.

PR #144, opened 2022-04-21
feat: Support Base64 encoding

soniqua:
#### What does this PR do?
This PR adds the capability to `base64` encode payloads created by the `snyk code test` with the `options.base64Encoding` flag.

This feature is added to allow `base64` encoded payloads to be passed to `deeproxy` for single-tenant solutions, obfuscating source code and preventing Akamai WAF rules incorrectly flagging code transmitted in the payload.

Enabling `base64` encoding requires `deeproxy` to be deployed *and* with the `base64` encode flag enabled also.

#### Ref
https://snyksec.atlassian.net/browse/NEBULA-520

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -25,6 +25,7 @@ export interface ConnectionOptions {
   sessionToken: string;
   source: string;
   requestId?: string;
+  base64Encoding: boolean;
 }
 
 // The trick to typecast union type alias
@@ -208,6 +209,13 @@ interface CreateBundleOptions extends ConnectionOptions {
 export async function createBundle(
   options: CreateBundleOptions,
 ): Promise<Result<RemoteBundle, CreateBundleErrorCodes>> {
+  let payloadBody;
+  if (options.base64Encoding) {
+    const payloadBuffer = Buffer.from(JSON.stringify(options.files));
+    payloadBody = payloadBuffer.toString('base64');
+  } else {
+    payloadBody = options.files;
+  }
   const payload: Payload = {
     headers: {
       ...prepareTokenHeaders(options.sessionToken),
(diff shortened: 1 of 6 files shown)


src/http.ts
```diff
@@ -208,6 +209,13 @@ interface CreateBundleOptions extends ConnectionOptions {
 export async function createBundle(
   options: CreateBundleOptions,
 ): Promise<Result<RemoteBundle, CreateBundleErrorCodes>> {
+  let payloadBody;
+  if (options.base64Encoding) {
+    const payloadBuffer = Buffer.from(JSON.stringify(options.files));
+    payloadBody = payloadBuffer.toString('base64');
```
metju90, line 215: Question: Do you think this could have a negative effect on performance? Like does it slow down file upload of big repos by 300%?

tests/analysis.spec.ts
```diff
@@ -119,6 +119,7 @@ describe('Functional test of analysis', () => {
           source,
           bundleHash: bundle.fileBundle.bundleHash,
           files: [],
+          base64Encoding: false
```
metju90, line 122: Would be cool to have test for ` base64Encoding: true` case

soniqua: Agreed - I don't think I can do this until the `deeproxy` changes are made live though.

metju90: Not sure how the structure of tests is here, but generally, for a unit test you can.

You can check if the request contains encoded payloads. We don't care what deeproxy response with (we just want to test the code changes of code-client)

src/http.ts
soniqua: There will be a negative impact on performance. I'll get some comparative tests and add here.

tests/analysis.spec.ts
soniqua: Unit tests added

src/http.ts
soniqua: As discussed - we'll get a separate activity raised to evaluate performance

metju90 (approved):

novalex (approved):

PR #160, opened 2022-09-20
fix: fast-glob perf issues [ROAD-1087]

michelkaporin:
- [ ] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
On some machines `@snyk/fast-glob` causes performance bottleneck. See [this ticket](https://snyksec.atlassian.net/browse/ROAD-1087). Our `@snyk/fast-glob` fork is quite outdated and not maintained. Thus I thought we can drop its dependency and replace with the official fast-glob lib.

Based on my tests replacement didn't result in any issues.

#### Additional questions
Wondering if we can completely ditch `@snyk/fast-glob`, since it's not that used according to my searches?
I saw it was introduced to [patch DDoS vulnerability](https://github.com/gulpjs/glob-parent/compare/main...snyk:glob-parent:main) in fast-glob dependency that has been addressed in the upstream (can't find a commit anymore though).


strassl-snyk (approved):
Well that's certainly an interesting issue. LGTM.

From what I can see I agree with your assessment that there is no reason to keep `@snyk/fast-glob`. It was basically introduced only for #59 and, while the exact reason has been lost to time, it seems to have been a temporary "we need to patch this library ourselves" kind of fix. If this has been upstreamed I see no reason to maintain our patch version and I also couldn't find any consumers within Snyk. Not sure who has access to archive that repo though.

michelkaporin:
@strassl-snyk I'll chat with folks to archive the repos. Before I do that just want to get final ✅ from the customer that it solves their perf problem with latest dependency wired in VS Code extension. In any case, makes sense to move to it in `code-client`.

michelkaporin:
@strassl-snyk I want to merge this in, but failing `default_workflow` & `security/snyk` prevent me from it.

The foremost seem to fail on integration tests for all opened PRs and is flaky for a while, as I was changing it quite many times already in my previous PRs in this year. 

I don't have permissions to rerun the latter but looks like this check was rate limited.

How should we proceed here?

michelkaporin:
@strassl-snyk I've fixed the tests but `Snyk AppSec` checks are still there and not even reporting a status 🤌

Can you perhaps just merge this PR as admin, bypassing these two last checks? Hopefully other PRs won't have them.

PR #161, opened 2022-09-21
feat: add org propagation [ZEN-340]

patricia-v:
Jira ticket: [ZEN-340](https://snyksec.atlassian.net/browse/ZEN-340)

### Related PRs
[deeproxy#211](https://github.com/snyk/deeproxy/pull/211)
[registry#29177](https://github.com/snyk/registry/pull/29177)

### Context
Since the introduction of the new services, we started fetching the org's FFs in `registry` when `deeproxy` performs a token validation to its endpoint. Previously, this operation did not require the current org, so the default one was used by `registry`.

We currently want the FFs of the org being used, so we need to propagate the org provided in `CLI` by either the flag or the config.

#### Relevant code-client requests to propagate the org
From my investigation, it seems that the only `code-client` requests performed by `CLI` that triggers a token verification are:

- createBundle()
- checkBundle()
- extendBundle()
- getAnalysis()

The last one, `getAnalysis()`, already had the proper relay of the org, so the only change necessary was adding the header for `deeproxy`.

The flows (and changes) for the requests are simplified in the following diagrams:

image

image

### TODO
- [X] Tests
- [ ] Update CLI to use latest version of code-client (after merge)


src/http.ts
```diff
@@ -246,6 +247,7 @@ export async function createBundle(
       source: options.source,
       ...(options.requestId && { 'snyk-request-id': options.requestId }),
       ...(base64Encoding ? { 'content-type': 'application/octet-stream', 'content-encoding': 'gzip' } : null),
+      ...(options.org ? { 'snyk-org': options.org } : {}),
```
novalex, line 250: `enh`: suggest using the simplified version like for `request-id`.

src/http.ts
```diff
@@ -283,6 +285,7 @@ export async function checkBundle(options: CheckBundleOptions): Promise<Result<R
       ...prepareTokenHeaders(options.sessionToken),
       source: options.source,
       ...(options.requestId && { 'snyk-request-id': options.requestId }),
+      ...(options.org ? { 'snyk-org': options.org } : {}),
```
novalex, line 288: `enh`: same as above

src/http.ts
```diff
@@ -331,6 +334,7 @@ export async function extendBundle(
       source: options.source,
       ...(options.requestId && { 'snyk-request-id': options.requestId }),
       ...(base64Encoding ? { 'content-type': 'application/octet-stream', 'content-encoding': 'gzip' } : null),
+      ...(options.org ? { 'snyk-org': options.org } : {}),
```
novalex, line 337: `enh`: same as above

src/http.ts
```diff
@@ -390,6 +394,7 @@ export async function getAnalysis(
       ...prepareTokenHeaders(options.sessionToken),
       source: options.source,
       ...(options.requestId && { 'snyk-request-id': options.requestId }),
+      ...(options.org ? { 'snyk-org': options.org } : {}),
```
novalex, line 397: `enh`: same as above

src/http.ts
novalex, line 250:
```suggestion
      ...(options.org && { 'snyk-org': options.org }),
```

novalex (approved):

PR #163, opened 2022-10-11
fix[ZEN-371, COD-841]: allow capitalised extensions

pkey:
#### What does this PR do?

While working on one of the customer tickets, I found that capitalised file extensions are not picked up by our files parser - which is not the case for the UI. After some [discussions](https://snyk.slack.com/archives/C0370TRPF7F/p1665399622050539) within the team, we have decided to align that functionality and respect capitalised file extensions.

Try a folder that contains files with, for example, a `.JS` extension. It shouldn't work with the previous version (no files will be found) but it should be fine with this change.

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -1,6 +1,6 @@
 import * as nodePath from 'path';
 import * as fs from 'fs';
-import fg from 'fast-glob';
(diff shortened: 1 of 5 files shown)


pkey:
closing since the priority is not clear

PR #167, opened 2023-01-09
[NEBULA-958] feat/upload dot snyk file

soniqua:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

- Allows the upload of `.snyk` files via `code-client` - note that the `/api/filters` endpoint must also be updated to allow this file to be uploaded.
- Adds the `jest-extended` matchers to allow matching of arrays of objects ignoring order.
- Fixes the CircleCI cache mechanism
- Replaces deprecated CircleCI convenience images

- Review changes to tests to accommodate the new file upload.

- Pull this branch and run `npm run test` locally.
- If testing with Code API (old flow) you'll need to update the filtering/file deletion to 

- This is required to enable custom rules.


src/files.ts
```diff
@@ -134,7 +135,7 @@ export function parseFileIgnores(path: string): string[] {
 export function getGlobPatterns(supportedFiles: SupportedFiles): string[] {
   return [
     ...supportedFiles.extensions.map(e => `*${e}`),
-    ...supportedFiles.configFiles.filter(e => !IGNORE_FILES_NAMES.includes(e)),
+    ...supportedFiles.configFiles.filter(e => !EXCLUDED_NAMES.includes(e)),
```
neonnoon, line 138:
```
export const EXCLUDED_NAMES = [GIT_FILENAME, GITIGNORE_FILENAME, DCIGNORE_FILENAME];
export const IGNORE_FILES_NAMES = [GITIGNORE_FILENAME, DCIGNORE_FILENAME, DOTSNYK_FILENAME];
```

This allows `.snyk` but now excludes `.git`. Interestingly, `.git` seems to already be excluded by [this](https://github.com/snyk/code-client/blob/master/src/constants.ts#L19). Still, there could be some unexpected side-effects?

neonnoon (changes requested):

src/files.ts
soniqua: Good catch. We only use `EXCLUDED_NAMES` once in this project (and I can't see it being used by any `code-client` consumers), so I can tweak the object.

soniqua: Done - tested locally, `.git` is indeed excluded elsewhere.

neonnoon (approved):

tests/analysis.spec.ts
```diff
@@ -88,14 +88,14 @@ describe('Functional test of analysis', () => {
         expect(new Set(bundle.analysisResults.coverage)).toEqual(
           new Set([
             {
-              files: 2,
```
strassl-snyk, line 91: q: Why reorder those? Since this is a set, I would expect it to be order-insensitive?

strassl-snyk (approved):
LGTM

enh: You have two `feat: upload .snyk file` commits (probably a mistake while commiting)?
enh: This PR seems to combine unrelated (?) changes (CircleCI, .snyk upload) into the same commit - might be easier to review and cleaner in general to break them into separate commits?

tests/analysis.spec.ts
soniqua: Good catch - I'll refactor these commits and remove this change.

PR #169, opened 2023-02-14
fix: fixed http library to properly set API header with organization …

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

fix: fixed http library to properly set API header with organization name

Currently, we don't set a header with organization name for getAnalysis call.
These changes are all internal and don't affect public API

### DIFF
--- a/src/analysis.ts
+++ b/src/analysis.ts
@@ -100,7 +100,6 @@ export async function analyzeFolders(options: FileAnalysisOptions): Promise<File
     ...options.connection,
     ...options.fileOptions,
     languages: options.languages,
-    ...(options.analysisContext ? { analysisContext: options.analysisContext } : {}),
   });
   if (fileBundle === null) return null;
 
--- a/src/bundles.ts
+++ b/src/bundles.ts
@@ -250,10 +250,9 @@ export async function createBundleFromFolders(options: CreateBundleFromFoldersOp
   }
 
   const bundleOptions = {
-    ...pick(options, ['baseURL', 'sessionToken', 'source', 'requestId']),
+    ...pick(options, ['baseURL', 'sessionToken', 'source', 'requestId', 'org']),
     baseDir,
     files: bundleFiles,
-    ...(options.analysisContext?.org?.name ? { org: options.analysisContext.org.name } : {}),
   };
 
   // Create remote bundle
--- a/src/http.ts
+++ b/src/http.ts
@@ -181,7 +181,7 @@ export async function getFilters(
   return generateError<GenericErrorTypes>(res.errorCode, GENERIC_ERROR_MESSAGES, apiName);
 }
 
-function prepareHeaders(options: ConnectionOptions) {
+function commonHttpHeaders(options: ConnectionOptions) {
   return {
     'Session-Token': options.sessionToken,
     // We need to be able to test code-client without deepcode locally
@@ -226,7 +226,7 @@ export async function createBundle(
     headers: {
       'content-type': 'application/octet-stream',
       'content-encoding': 'gzip',
-      ...prepareHeaders(options),
+      ...commonHttpHeaders(options),
     },
     url: `${options.baseURL}/bundle`,
     method: 'post',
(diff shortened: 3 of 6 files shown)


src/http.ts
```diff
@@ -363,10 +357,8 @@ export async function getAnalysis(
 ): Promise<Result<GetAnalysisResponseDto, GetAnalysisErrorCodes>> {
   const config: Payload = {
     headers: {
-      ...prepareTokenHeaders(options.sessionToken),
-      source: options.source,
-      ...(options.requestId && { 'snyk-request-id': options.requestId }),
-      ...(options.org && { 'snyk-org-name': options.org }),
+      ...prepareHeaders(options),
+      ...(options.analysisContext?.org?.name && { 'snyk-org-name': options.analysisContext.org.name }),
```
strassl-snyk, line 361: q: Why the duplication here? We are pulling the org name from `options.org` but then also overriding it with the `analysisContext`? Why are we doing this here but not for the `createBundle/extendBundle` calls?

strassl-snyk (commented):
Could someone explain to me (in very simple words 😅) what the previous behavior was, why it was incorrect, and what exactly we expect to change?

cc @patricia-v this will have to be integrated into our code-client changes for `/report`

src/http.ts
Arvi3d: this is a mess, i agree, but i'm trying to minimise my refactoring, while still achieve goals.
So, the answer is createBundle/extendBundle does not accept analysisContext.
Maybe it would make sense to add it there, but it's a call I leave to you folks later on.

Arvi3d: let me do a bit more refactoring, as this is a point of friction now.
I'll make org available only for bundle requests. to remove duplication

strassl-snyk: I'm still not sure I understand - under what circumstances would the org name not be passed as `options.org` but included in `options.analysisContext.org.name`?

Arvi3d: pretty much this is only way it works now. So, now externally only analysisContext is used for "analyzeFolders" function call. noone uses org.
Then, org is only injected internally for bundle creating: https://github.com/snyk/code-client/blob/master/src/bundles.ts#L256
I did not touch this part in this PR.

Arvi3d: let me clean this code and i'll ask you to re-review a cleaner version

Arvi3d: @strassl-snyk pls re-review

src/bundles.ts
```diff
@@ -250,10 +250,9 @@ export async function createBundleFromFolders(options: CreateBundleFromFoldersOp
   }
 
   const bundleOptions = {
-    ...pick(options, ['baseURL', 'sessionToken', 'source', 'requestId']),
+    ...pick(options, ['baseURL', 'sessionToken', 'source', 'requestId', 'org']),
     baseDir,
     files: bundleFiles,
-    ...(options.analysisContext?.org?.name ? { org: options.analysisContext.org.name } : {}),
```
strassl-snyk, line 256: See above

src/analysis.ts
```diff
@@ -100,7 +100,6 @@ export async function analyzeFolders(options: FileAnalysisOptions): Promise<File
     ...options.connection,
     ...options.fileOptions,
     languages: options.languages,
-    ...(options.analysisContext ? { analysisContext: options.analysisContext } : {}),
```
strassl-snyk, line 103: q: I assume the analysisContext is never used in this flow?

tests/analysis.spec.ts
```diff
@@ -301,30 +301,50 @@ describe('Functional test of analysis', () => {
...
+            name: 'proj',
+            publicId: 'id',
+            type: 'code',
+          },
         },
       };
 
       const makeRequestSpy = jest.spyOn(needle, 'makeRequest');
 
-      await analyzeFolders({
```
strassl-snyk, line 313: q: Why change this from `analyzeFolders` to `analyzeBundle`?

strassl-snyk (approved):
Changes look okay to me, but it's very hard to tell given how convoluted the control flow in code-client is. Assuming this has been tested and works as intended we should be good.

So next steps:
- Pass `options.org` in the CLI so it actually works.

src/analysis.ts
Arvi3d: > q: I assume the analysisContext is never used in this flow?

yes

tests/analysis.spec.ts
Arvi3d: otherwise this test will not be able to call "/analysis", as it will break at bundle creation step.
Since i did not want to change the purpose of the test, this was the way.
Btw, let me add a new test for /bundle calls only

PR #174, opened 2023-02-21
feat: added an optional parameter for public user identificator

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

This is also a continuation of changes required to make deeproxy's "systemcheck" call properly communicate with backend services.

This PR adds an optional parameter for public user id. This is required as our files-bundle-store requires it: https://github.com/snyk/files-bundle-store/blob/main/src/api/rest/middleware/auth.ts#L8

related jira ticket: https://snyksec.atlassian.net/browse/ZEN-529

### DIFF
--- a/src/bundles.ts
+++ b/src/bundles.ts
@@ -47,7 +47,7 @@ async function* prepareRemoteBundle(
   emitter.createBundleProgress(cumulativeProgress, options.files.length);
   for (const chunkedFiles of composeFilePayloads(options.files, MAX_PAYLOAD)) {
     const apiParams = {
-      ...pick(options, ['baseURL', 'sessionToken', 'source', 'removedFiles', 'requestId', 'org']),
+      ...pick(options, ['baseURL', 'sessionToken', 'source', 'extraHeaders', 'removedFiles', 'requestId', 'org']),
       files: chunkedFiles.reduce((d, f) => {
         // deepcode ignore PrototypePollution: FP this is an internal code
         d[f.bundlePath] = f.hash;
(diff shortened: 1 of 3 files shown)


strassl-snyk (commented):
I'm not sure about this - at this point, we seem to be making non-trivial changes only to support an internal test setup, which probably should be built differently anyway. 

Case in point, there is never a reason why the `X-PublicUserId` should be passed from the client in a real environment. It's a trusted authentication header that is only set by the ingress proxy. It won't cause any harm to set it here since it will be overridden, but maybe we should reconsider just changing the tests?
Also, didn't the old flow also require this header to be set for bundle access authorization?

Arvi3d:
nope. old flow did not require that.
Alternatively, we can remove this authorization step inside files-bundle-store. should be harmless.
Alternatively, we can allow "extraHeaders" in a code-client, instead of specific publicuserid. But it does not feel good eighter.

novalex:
Agree with Simon about the necessity of passing this from the client - there is a real reason for passing org, as a user can belong to many, but the user ID should always be inferred upstream from the token/auth being used by the client.

Can the systemcheck not send the header itself? Why are these client changes necessary to make an automated test work?

(Sorry for likely a lot of missing context on my part, trying to understand the requirements and implications)

Arvi3d:
changed approach to use `extraHeaders` optional parameter. Folks please re-review

strassl-snyk (approved):
Not excited about doing this, but given the alternatives LGTM

PR #170, opened 2023-02-17
feat: add report upload to code test [ZEN-494]

patricia-v:
- [x] Ready for review
- [x] Linked to Jira ticket - [ZEN-494](https://snyksec.atlassian.net/browse/ZEN-494)

#### What does this PR do?

- Adds the report feature to code-client
- Adds tests to the report feature

Related: [cli PR](https://github.com/snyk/cli/pull/4422)

This `code-client` version been successfully tested with the current `cli` version to confirm no breaking changes were made.

#### TODO ⚠️ 

- [ ] Finish tests when [ZEN-495](https://snyksec.atlassian.net/browse/ZEN-495) gets merged

[ZEN-494]: https://snyksec.atlassian.net/browse/ZEN-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ZEN-495]: https://snyksec.atlassian.net/browse/ZEN-495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #175, opened 2023-03-08
fix: add report tests [ZEN-537]

patricia-v:
- [x] Ready for review
- [x] Linked to Jira ticket - [ZEN-537](https://snyksec.atlassian.net/browse/ZEN-537)

#### What does this PR do?

Adds missing report tests from #170

[ZEN-537]: https://snyksec.atlassian.net/browse/ZEN-537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/api.spec.ts
+++ b/tests/api.spec.ts
@@ -10,6 +10,7 @@ import {
   getAnalysis,
   AnalysisStatus,
   initReport,
+  getReport,
 } from '../src/http';
 import { BundleFiles } from '../src/interfaces/files.interface';
 import * as needle from '../src/needle';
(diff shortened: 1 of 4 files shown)


tests/api.spec.ts
```diff
@@ -382,49 +383,4 @@ describe('Requests to public API', () => {
     },
     TEST_TIMEOUT,
   );
-
-  // TODO: this test is being skipped for now since the /report flow hasn't been fully rolled out and it can't succeed for now
-  it.skip('test successful report workflow', async () => {
```
patricia-v, line 387: This test didn't seem to make much sense based on the others we already have, so I decided to remove it, but it can be discussed

strassl-snyk (approved):
LGTM


Do we by any chance document the test org setup procedure somewhere in case we need to recreate it? I assume you had to enable the FF for it to work?

patricia-v:
> Do we by any chance document the test org setup procedure somewhere in case we need to recreate it? I assume you had to enable the FF for it to work?

@strassl-snyk I'm not certain if we have it documented, but I didn't turn on the FF because I mocked the return values of the tests that were sending the `400` as the results weren't needed from deeproxy in that case. Also, I'm not sure if it makes sense to have tests in `code-client` that depend on FFs, I think that should be something done in `deeproxy`/`registry`, what do you think?

About the account used in the tests, I'm fairly certain that it is in 1Password, at the least the `SNYK_API_KEY` used in the CircleCI env variable matches it for `app.dev.snyk.io` (just double checked it)

PR #176, opened 2023-03-15
feat: added CLI

Arvi3d:
- [ ] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

During my load tests i was struggling due to additional configurations required when used Code client via Snyk CLI.
Especially, if I wanted to adjust code-client code a bit and see it's affect on performance/stability.
Also, i wanted to have a isolated functionality to only create bundle, without running analysis on it.

This PR adds a CLI, that allows to do so.

Example usage:
```bash
$ time npm run cli -- bundle:create --url="<deeproxy url or directly files-bundle-store>" --token="snyk token" --headers="custom-header: test-user" --patterns=".*" --path="<path to source code>"
```

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -431,7 +431,8 @@ export function resolveBundleFilePath(baseDir: string, bundleFilePath: string): 
   return decodeURI(relPath);
 }
 
-export function* composeFilePayloads(files: FileInfo[], bucketSize = MAX_PAYLOAD): Generator<FileInfo[]> {
+// MAX_PAYLOAD / 2 is because every char takes 2 bytes in the payload
+export function* composeFilePayloads(files: FileInfo[], bucketSize = MAX_PAYLOAD / 2): Generator<FileInfo[]> {
   type Bucket = {
     size: number;
     files: FileInfo[];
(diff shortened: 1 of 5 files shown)


src/files.ts
```diff
@@ -431,7 +431,8 @@ export function resolveBundleFilePath(baseDir: string, bundleFilePath: string):
   return decodeURI(relPath);
 }
 
-export function* composeFilePayloads(files: FileInfo[], bucketSize = MAX_PAYLOAD): Generator<FileInfo[]> {
+// MAX_PAYLOAD / 2 is because every char takes 2 bytes in the payload
+export function* composeFilePayloads(files: FileInfo[], bucketSize = MAX_PAYLOAD / 2): Generator<FileInfo[]> {
```
strassl-snyk, line 435: q: Do the changes in this function have an impact on production functionality?

Arvi3d: only on client. this function is responsible of creating chunks of data, that can fit in one request. As a result of these changes, it will actually create smaller chunks. But won't have an impact. Code-client is used now only in CLI and old version of VSCe. Both will work well. Potentially, it can cause slower rate of download of initial project files. But based on my load tests, adjustments of payload size give max up to 10% difference in duration. In this can, it will be even less, thus unnoticeable

strassl-snyk: I see, thanks for the explanation. So what's the motivation for this change then if it causes a (slightly) worse UX?

Rest of the PR LGTM btw

Arvi3d: This change makes the calculation more correct and predictable. Less likely to extend beyond allowed limits. So, i believe more accurate calculation is worth unnoticeably worsen performance.

PR #177, opened 2023-03-22
fix: add support for additional report parameters [ZEN-556]

novalex:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

Add support and handling for additional report-related params:
- `targetName` (from CLI: `target-name`)
- `targetRef` (from CLI: `target-reference`)
- `remoteRepoUrl` (from CLI: `remote-repo-url`)

CLI companion PR: https://github.com/snyk/cli/pull/4484

TODO:
- [ ] Merge `code-orchestrator` companion PR: https://github.com/snyk/code-orchestrator/pull/40

### DIFF
--- a/tests/report.spec.ts
+++ b/tests/report.spec.ts
@@ -1,11 +1,11 @@
 import path from 'path';
-import { createBundleFromFolders } from '../src/bundles';
 import { baseURL, sessionToken, source } from './constants/base';
 import { reportBundle } from '../src/report';
 import * as http from '../src/http';
 import { sampleProjectPath, initReportReturn, getReportReturn } from './constants/sample';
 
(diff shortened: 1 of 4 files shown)


tests/report.spec.ts
patricia-v, line 1: q: do you think it's worth adding more failing tests here, or do you think the validation one in registry is enough?

patricia-v (approved):
LGTM!

tests/report.spec.ts
novalex: IMO more unhappy path tests would be great to have, but probably best to add them once we have the error handling/propagation in a better place, since it's not totally clear what the error return types would be.

PR #181, opened 2023-04-28
fix: Send valid `Authorization` header

florinMiro:
- [x] Ready for review

#### What does this PR do?

I am changing the AuthN logic to only send the token through the `Authorization` header instead of sending both `session-token` and a dummy `Authorization`.
This change relies on this [Deeproxy PR](https://github.com/snyk/deeproxy/pull/336) 

In the Deeproxy PR, there is more context on this AuthN change and a diagram of all Deeproxy consumers and how they are affected by this change.

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -184,9 +184,7 @@ export async function getFilters(
 
 function commonHttpHeaders(options: ConnectionOptions) {
   return {
-    'Session-Token': options.sessionToken,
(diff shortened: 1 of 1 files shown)


Arvi3d:
@florinMiro please ask folks from Zenith team to review this. I don't feel like i should be reviewing this repo.

strassl-snyk (approved):

PR #184, opened 2023-05-17
fix: case-sensitive file extensions filtering [ZEN-371] [ZEN-614]

michelkaporin:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?
Ignores file extension casing when filtering out for supported files before uploading for analysis.
Current tests are failing due to https://github.com/snyk/sast-analysis-api/pull/328

Build the lib with CLI, then run a CLI scan on any folder with a single file having an uppercase file extension. This file should not be filtered out for analysis, providing Code results back.

We're already case insensitive for file extensions around analysis in IDE. This PR addresses CLI. Next is to be case insensitive within Git and PR checks flow.


src/bundles.ts
```diff
@@ -209,12 +209,14 @@ export async function getSupportedFiles(
   const supportedFilesFromApi = resp.value;
   //Given supported languages from 'registy'
   if (languages) {
-    const supportedFiles: SupportedFiles = {} as any;
+    const supportedFiles = {} as SupportedFiles;
```
strassl-snyk, line 212: enh: Couldn't we initialize this directly as:

```ts
const supportedFiles: SupportedFiles = {
  configFiles: supportedFilesFromApi.configFiles,
  extensions: languages
}
```

This would improve the type safety, simplify the logic, and remove the cast.

src/bundles.ts
```diff
@@ -209,12 +209,14 @@ export async function getSupportedFiles(
...
     supportedFiles.configFiles = supportedFilesFromApi.configFiles;
     supportedFiles.extensions = languages;
+
     //For verification only
+    const userSupportedExtensions = supportedFilesFromApi.extensions.map(e => e.toLowerCase());
```
strassl-snyk, line 217: enh: A comment explaining the reasoning here might be helpful for future maintainers.

src/files.ts
```diff
@@ -134,11 +134,31 @@ export function parseFileIgnores(path: string): string[] {
 
 export function getGlobPatterns(supportedFiles: SupportedFiles): string[] {
   return [
-    ...supportedFiles.extensions.map(e => `*${e}`),
+    ...supportedFiles.extensions.map(e => `*${generateAllCaseGlobPattern(e)}`),
     ...supportedFiles.configFiles.filter(e => !EXCLUDED_NAMES.includes(e)),
   ];
 }
 
+// Generates glob patterns for case-insensitive file extension matching.
```
strassl-snyk, line 142: I love the explanations

src/files.ts
```diff
@@ -134,11 +134,31 @@ export function parseFileIgnores(path: string): string[] {
...
+// E.g. *.[jJ][sS] for matching .js files without case-sensitivity.
+function generateAllCaseGlobPattern(fileExtension: string): string {
+  const chars = Array.from(fileExtension);
+  const p = chars.reduce((pattern: string[], extensionChar, i) => {
+    if (i == 0) {
+      // first char is always '.', no need to generate multiple cases for file extension character
+      if (extensionChar != '.') {
+        throw new Error('Unexpected file extension pattern when constructing glob patterns.');
```
strassl-snyk, line 150: req/q: This will cause every analysis to fail if we ever return an entry that doesn't start with `.`, right? Maybe we should remove the special handling for the `.` character, uppercase what we can, and ignore what we cannot uppercase?

src/files.ts
```diff
@@ -134,11 +134,31 @@ export function parseFileIgnores(path: string): string[] {
...
+    ...supportedFiles.extensions.map(e => `*${generateAllCaseGlobPattern(e)}`),
     ...supportedFiles.configFiles.filter(e => !EXCLUDED_NAMES.includes(e)),
   ];
 }
 
+// Generates glob patterns for case-insensitive file extension matching.
+// E.g. *.[jJ][sS] for matching .js files without case-sensitivity.
+function generateAllCaseGlobPattern(fileExtension: string): string {
+  const chars = Array.from(fileExtension);
+  const p = chars.reduce((pattern: string[], extensionChar, i) => {
```
strassl-snyk, line 146: enh: `caseInsensitivePatterns` instead of `p` might be more descriptive?

src/files.ts
```diff
@@ -134,11 +134,31 @@ export function parseFileIgnores(path: string): string[] {
...
+    if (i == 0) {
+      // first char is always '.', no need to generate multiple cases for file extension character
+      if (extensionChar != '.') {
+        throw new Error('Unexpected file extension pattern when constructing glob patterns.');
+      }
+
+      return [extensionChar];
+    }
+
+    const regexCharPattern = `[${extensionChar.toLowerCase()}${extensionChar.toUpperCase()}]`;
```
strassl-snyk, line 156: q: Are we dealing with regexes or globs here? Asking because, while they may look similar, they don't have the same semantics (e.g. regarding `.`).

src/files.ts
```diff
@@ -134,11 +134,31 @@ export function parseFileIgnores(path: string): string[] {
 
 export function getGlobPatterns(supportedFiles: SupportedFiles): string[] {
   return [
-    ...supportedFiles.extensions.map(e => `*${e}`),
+    ...supportedFiles.extensions.map(e => `*${generateAllCaseGlobPattern(e)}`),
     ...supportedFiles.configFiles.filter(e => !EXCLUDED_NAMES.includes(e)),
```
strassl-snyk, line 138: q: Should these be case sensitive or case insensitive?

tests/files.spec.ts
```diff
@@ -148,4 +149,17 @@ describe('files', () => {
     const windowsPath = 'C:\\Users\\user\\Git\\goof%20test\\index.js';
     expect(resolveBundleFilePath(baseDir, windowsPath)).toEqual(decodeURI(windowsPath));
   });
+
+  it('generates correct glob patterns for supported files', () => {
```
strassl-snyk, line 153: enh: It would be useful to have tests for some of the weirder edge-cases, for example:
- Empty extensions
- Extensions not starting with a `.`
- Extensions with characters that don't have an uppercase variant

novalex (approved):
Nice work! 👌 

Agree with Simon's suggestions though, especially re. throwing the error for the first char.

src/files.ts
michelkaporin: Correct. The problem with ignoring a missing `.` is that we won't catch the problem in any way, keeping us running in a blind-mode. 

Ideally this validation should be somewhere on the backend actually, and client-side/analysis shouldn't be impacted. I'll convert this to `console.log` for general visibility.

src/files.ts
michelkaporin: Glob allows for a subset of regex syntax in [advanced pattern matching](https://github.com/mrmlnc/fast-glob#pattern-syntax). I'll rename the var to avoid confusion 👍

src/files.ts
michelkaporin: This is a good question. It's case sensitive currently. Making these case insensitive would be a different problem to solve and out of this PR scope, since the change would be not only in `code-client`.

tests/files.spec.ts
michelkaporin: Good point, although we shouldn't be returning these cases :)

strassl-snyk (approved):

PR #189, opened 2023-07-13
fix: file tests after fast-glob update

novalex:
- [x] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

Following a minor update to `fast-glob` (version `3.3.0`), we are seeing different behaviour for the `searchFiles` function.

The new behaviour is actually the correct one, as there was no reason for `exclude/.snyk` to not have been detected beforehand, so the update seems correct and the tests were updated to reflect the new behaviour.

Furthermore, instead of checking an individual file in the `collect bundle files` test we now check all of the items returned against a snapshot. This will make it much clearer when/if anything changes between the snapshot and actual results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@types/write": "^2.0.1",
     "@typescript-eslint/eslint-plugin": "^4.31.2",
     "@typescript-eslint/parser": "^4.0.1",
+    "commander": "^10.0.0",
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.24.2",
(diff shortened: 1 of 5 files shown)


vasser (approved):

package.json
```diff
@@ -79,7 +79,7 @@
     "@types/lodash.union": "^4.6.6",
     "@types/sarif": "^2.1.4",
     "@types/uuid": "^8.3.1",
-    "fast-glob": "^3.2.11",
+    "fast-glob": "3.3.0",
```
michelkaporin, line 82: Why not to use the latest minor?
P.S. you should lock the package.json

novalex: Because the latest minor is what introduced test failures with no code changes, and because the lock file is not commited (see https://github.com/snyk/code-client/pull/43).

novalex: True, but we should have ensured correct behaviour beforehand, rather than rely on arbitrary package updates to fix our logic for us 😅 
There's also the question of "was it really a minor update", since obviously it introduced potentially breaking changes, but that's a separate discussion, probably for the maintainers of `fast-glob`.
But it seemed sensible to lock this version that we know is correct, and we can update as needed, when new features or security fixes come up.

> If I understand correctly, it's to ensure arbitrary dep graph consistency between our tests and CLI when being published.

That is my understanding too, we want to keep our dev envs and pipelines in line with envs using the actual library which won't have the lock file.

PR #192, opened 2023-07-25
feat: lock library dependencies

michelkaporin:
- [x] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?
This PR locks code-client dependencies for install reproducibility purposes in the CI. 

The current solution of not having a lock, doesn't address the issue that [it's meant to solve](https://github.com/snyk/code-client/pull/43), as CLI (or any other consumer) would result in a different dependency resolution than the one happening in the current library's CI pipeline.

Coffee read: https://snyk.io/blog/what-is-package-lock-json. 
Discussions: https://snyk.slack.com/archives/C0135Q8LP61/p1689183248655649


strassl-snyk (approved):
Changes LGTM, Snyk seems unhappy though.

I do wonder whether it would be worthwhile to have an additional CI step that tests against latest. We could even run that on a schedule. This way we could reliably distinguish between a dependency breaking and PR changes breaking. I don't want to blow up the scope for this change though.

.circleci/config.yml
```diff
@@ -8,22 +8,34 @@ jobs:
...
+                command: npm install
+      - unless:
+          condition: << parameters.package-lock >>
+          steps:
+            - run:
+                name: Install without locked dependencies
+                command: npm install --no-package-lock
       - save_cache:
-          key: v2-{{ .Branch }}-{{ .Revision }}-<< parameters.node_version >>
+          key: v2-{{ .Branch }}-{{ .Revision }}-<< parameters.node_version >>-{{ checksum "package-lock.json" }}
```
strassl-snyk, line 34: q: Is this correct when running with `--no-package-lock` or would this pollute the cache?

michelkaporin:
@strassl-snyk great point, added it not to regress on the “surprise” end that we had so far covered as part of the existing CI.

strassl-snyk (changes requested):

.circleci/config.yml
michelkaporin: This will pollute the cache. I'll substitute checksum with `package-lock` parameter for cache persistency.

strassl-snyk (approved):

PR #199, opened 2023-08-02
fix: descriptive error when .snyk contains invalid ignores

vasser:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

Makes an error that is thrown in case of invalid ignore rules in .snyk more informative 

https://snyksec.atlassian.net/browse/ZEN-733

Run snyk code test -d over [Snyk CLI](https://github.com/snyk/cli) repository

https://snyk.slack.com/archives/C01DGQ3AT09/p1690922427417839?thread_ts=1689170972.097249&cid=C01DGQ3AT09

#### Screenshots

Screenshot 2023-08-02 at 12 09 41

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -132,7 +132,7 @@ export function parseFileIgnores(path: string): string[] {
   try {
     return parseIgnoreRulesToGlobs(rules, dirname);
   } catch (err) {
-    console.error('Could not parse ignore rules to glob', { path, error: (err as Error).message });
+    console.error('Could not parse ignore rules to glob', { path });
     throw new Error('Ignore rules should be in .gitignore format');
   }
 }
--- a/src/files.ts
+++ b/src/files.ts
@@ -133,7 +133,7 @@ export function parseFileIgnores(path: string): string[] {
     return parseIgnoreRulesToGlobs(rules, dirname);
   } catch (err) {
     console.error('Could not parse ignore rules to glob', { path });
-    throw new Error('Ignore rules should be in .gitignore format');
+    throw new Error('Please make sure ignore file follows correct syntax');
   }
 }
 
--- a/src/files.ts
+++ b/src/files.ts
@@ -129,7 +129,12 @@ export function parseFileIgnores(path: string): string[] {
       );
     }
   }
-  return parseIgnoreRulesToGlobs(rules, dirname);
+  try {
+    return parseIgnoreRulesToGlobs(rules, dirname);
+  } catch (err) {
+    console.error('Could not parse ignore rules to glob', { path, error: (err as Error).message });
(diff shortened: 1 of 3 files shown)


src/files.ts
```diff
@@ -129,7 +129,12 @@ export function parseFileIgnores(path: string): string[] {
       );
     }
   }
-  return parseIgnoreRulesToGlobs(rules, dirname);
+  try {
+    return parseIgnoreRulesToGlobs(rules, dirname);
+  } catch (err) {
+    console.error('Could not parse ignore rules to glob', { path, error: (err as Error).message });
+    throw new Error('Ignore rules should be in .gitignore format');
```
michelkaporin, line 136: The problem was occurring in `.snyk` file, which doesn't follow `.gitignore` format.

src/files.ts
```diff
@@ -129,7 +129,12 @@ export function parseFileIgnores(path: string): string[] {
       );
     }
   }
-  return parseIgnoreRulesToGlobs(rules, dirname);
+  try {
+    return parseIgnoreRulesToGlobs(rules, dirname);
+  } catch (err) {
+    console.error('Could not parse ignore rules to glob', { path, error: (err as Error).message });
```
michelkaporin, line 135: Is revealing an internal error message needed for a CLI user?

src/files.ts
michelkaporin, line 136: Can we terminate execution with an error instead of throwing, or is throwing a common approach at the moment?

vasser: @michelkaporin The problem can actually happen any ignore file. Not .snyk only. This is why I added path to the console error, so end user will be able to see where the problematic rule was defined

vasser: As far as I can see that is the common approach

novalex: IIRC when a flow throws in the CLI it automatically records error metrics (in Registry), which we might not want in case of invalid ignore files. But also IIRC changing the high-level CLI flow/terminating early is not trivial, so not a blocker if it provides more context and allows users to fix the problem.

src/files.ts
```diff
@@ -129,7 +129,12 @@ export function parseFileIgnores(path: string): string[] {
       );
     }
   }
-  return parseIgnoreRulesToGlobs(rules, dirname);
+  try {
+    return parseIgnoreRulesToGlobs(rules, dirname);
+  } catch (err) {
+    console.error('Could not parse ignore rules to glob', { path });
+    throw new Error('Please make sure ignore file follows correct syntax');
```
michelkaporin, line 136: grammar nazi nit: every sentence should end with a dot (here and above). disclaimer: not a native speaker here 😄

michelkaporin (approved):
user friendliness 👍

PR #200, opened 2023-08-18
fix: file exclude behaviour [ZEN-588]

novalex:
- [x] Ready for review
- [x] Linked to Jira ticket

#### What does this PR do?

Addresses logic bugs in the behaviour of file excludes and ignore rules.

Current (incorrect) behaviour:
- it collects all exclude/ignore rules from all supported files in the directory, including from directories excluded by the root-level .snyk file
	- i.e. if a directory `sub` is excluded by the top-level `.snyk` file, we still traverse it and collect any supported ignore files, which is unexpected and wrong behaviour - excluded directories should not be traversed at all.
- it applies negative match rules, which in some cases can override the .snyk exclude rules
	- i.e. if a `.snyk` exclude is defined for a directory or file `sub`, and there is a negative match for the same directory or file in the `.gitignore` file (`!sub`), `sub` will not be excluded as expected.

In short, these fixes aim to more strictly enforce file and directory exclusions defined in the `.snyk` file.

The way it works post fixes:
- Collect exclusion rules from all `.snyk` files in directory, if present
- Collect ignore rules from all other supported ignore files (`.dcignore`, `.gitignore`), except the ones found in paths excluded from the root `.snyk` file
- Return separate sets of rules for excludes and ignores
- Apply combined positive ignore rules to first supported files collection pass
- Apply only exclude rules to second file pass for negative pattern collection

This ensures that `.snyk` exclude rules always get apply, and cannot be overridden by negative ignore rules.

Performing a simple benchmark with 25 iterations of ignore rule collection against Registry, I did not find any significant performance difference between the old and new implementations.
As all local benchmarks it should be taken with a grain of salt, however I do not expect any noticeable slowdown of the collection, and depending on the number of files and exclude rules, customers might see a performance improvement as we no longer collect and parse ignore rules from files in excluded directories.

```
describe('performance benchmark', () => {
  it.each([
    ['new implementation', collectFilePolicies],
    ['old implementation', collectIgnoreRules],
  ])(`%s`, async (implementationType, collectIgnoreRulesFn) => {
    const collectPath = '/Users/alex/Projects/registry';

    const timings: number[] = [];

    for (let i = 0; i < 25; i++) {
      const t1 = Date.now();
      await collectIgnoreRulesFn([collectPath]);
      const t2 = Date.now();
      const time = t2 - t1;

      timings.push(time);
    }

    const average = timings.reduce((a, b) => a + b, 0) / timings.length;
    console.log(`Average time for ${implementationType}: ${average} milliseconds.`);
  });
});
```

Results:
```
  performance benchmark
    ✓ new implementation (31313 ms)
    ✓ old implementation (29438 ms)

Average time for new implementation: 1252.12 milliseconds.
Average time for old implementation: 1177.36 milliseconds.
```

https://snyksec.atlassian.net/browse/ZEN-588
https://snyksec.atlassian.net/browse/ZEN-630
https://snyksec.atlassian.net/browse/ZEN-773

### DIFF
--- a/src/files.ts
+++ b/src/files.ts
@@ -249,14 +249,10 @@ async function* searchFiles(
   patterns: string[],
   cwd: string,
   symlinksEnabled: boolean,
-  policies?: FilePolicies,
+  policies: FilePolicies,
 ): AsyncGenerator<string | Buffer> {
-  const positiveIgnores = policies
-    ? [...policies.excludes, ...policies.ignores.filter(rule => !rule.startsWith('!'))]
-    : [];
-  const negativeIgnores = policies
-    ? policies.ignores.filter(rule => rule.startsWith('!')).map(rule => rule.substring(1))
-    : [];
+  const positiveIgnores = [...policies.excludes, ...policies.ignores.filter(rule => !rule.startsWith('!'))];
+  const negativeIgnores = policies.ignores.filter(rule => rule.startsWith('!')).map(rule => rule.substring(1));
   // We need to use the ignore rules directly in the stream. Otherwise we would expand all the branches of the file system
   // that should be ignored, leading to performance issues (the parser would look stuck while analyzing each ignored file).
   // However, fast-glob doesn't address the negative rules in the ignore option correctly.
@@ -287,7 +283,7 @@ async function* searchFiles(
       followSymbolicLinks: symlinksEnabled,
       baseNameMatch: false,
       // Exclude rules should still be respected
-      ignore: policies?.excludes ?? [],
+      ignore: policies.excludes,
     });
     for await (const filePath of negativeSearcher) {
       if (isMatch(filePath.toString(), deepPatterns)) yield filePath;
--- a/src/files.ts
+++ b/src/files.ts
@@ -202,10 +202,12 @@ export async function collectIgnoreRules(
   const customRules = await Promise.all(tasks);
 
   const rules = union(fileIgnores, ...customRules);
-  const deduplicatedRules: string[] = [];
 
+  // Deduplicate rules, such as "**/dir" and "sub/dir".
+  // In this case "sub/dir" is a subset of the first rule and thus redundant.
+  const deduplicatedRules: string[] = [];
   for (const rule of rules) {
-    if (!deduplicatedRules.some(existingRule => multimatch(rule, [existingRule]).length > 0)) {
+    if (!deduplicatedRules.some(existingRule => multimatch(rule, existingRule).length > 0)) {
       deduplicatedRules.push(rule);
     }
   }
--- a/src/files.ts
+++ b/src/files.ts
@@ -201,23 +201,35 @@ export async function collectIgnoreRules(
         ignore: topLevelSnykIgnores,
       },
     );
-    // Read ignore files and merge new patterns
-    return union(...localIgnoreFiles.map(parseFileIgnores));
+
+    // Read ignore files and merge patterns.
+    const localIgnoreRules = union(...localIgnoreFiles.map(parseFileIgnores));
+
+    // Final array of rules after processing to return.
+    const rulesAfterProcessing: string[] = [];
+
+    for (const rule of localIgnoreRules) {
+      // Identify and remove negative rules that are superseded by top-level .snyk rules.
+      // i.e. if a directory "foo" is ignored by a top-level .snyk rule, then a negative match
+      // for "!foo" from another ignore file is superseded and needs to be removed.
+      const negativeRuleIsSupersededBySnykExcludes =
+        rule.startsWith('!') &&
+        topLevelSnykIgnores.some(existingRule => multimatch(rule.slice(1), existingRule).length > 0);
+      // Deduplicate rules, such as "**/dir" and "sub/dir".
+      // In this case "sub/dir" is a subset of the first rule and thus redundant.
+      const ruleIsRedundant = rulesAfterProcessing.some(existingRule => multimatch(rule, existingRule).length > 0);
+
+      if (!negativeRuleIsSupersededBySnykExcludes && !ruleIsRedundant) {
+        rulesAfterProcessing.push(rule);
+      }
+    }
+
+    return rulesAfterProcessing;
   });
+
   const customRules = await Promise.all(tasks);
 
-  const rules = union(fileIgnores, ...customRules);
-
-  // Deduplicate rules, such as "**/dir" and "sub/dir".
-  // In this case "sub/dir" is a subset of the first rule and thus redundant.
-  const deduplicatedRules: string[] = [];
-  for (const rule of rules) {
-    if (!deduplicatedRules.some(existingRule => multimatch(rule, existingRule).length > 0)) {
-      deduplicatedRules.push(rule);
-    }
-  }
-
-  return deduplicatedRules;
+  return union(fileIgnores, ...customRules);
 }
 
 export function determineBaseDir(paths: string[]): string {
--- a/src/files.ts
+++ b/src/files.ts
@@ -18,7 +18,7 @@ import {
   DOTSNYK_FILENAME,
   EXCLUDED_NAMES,
 } from './constants';
-import { CollectBundleFilesOptions } from './interfaces/analysis-options.interface';
+import { CollectBundleFilesOptions, FilePolicies } from './interfaces/analysis-options.interface';
 import { SupportedFiles, FileInfo } from './interfaces/files.interface';
 
 const isWindows = nodePath.sep === '\\';
@@ -178,16 +178,21 @@ function generateAllCaseGlobPattern(fileExtension: string): string {
 }
 
 /**
- * Recursively collect all exclude and ignore rules from directory.
+ * Recursively collect all exclude and ignore rules from "dirs".
  *
  * Exclude rules from .snyk files and ignore rules from .[*]ignore files are collected separately.
  * Any .[*]ignore files in paths excluded by .snyk exclude rules are ignored.
  */
-export async function collectIgnoreRules(
+export async function collectFilePolicies(
   dirs: string[],
   symlinksEnabled = false,
   fileIgnores: string[] = IGNORES_DEFAULT,
-): Promise<string[]> {
+): Promise<FilePolicies> {
+  const excludes: Set<string> = new Set();
+  const ignores: Set<string> = new Set();
+
+  fileIgnores.map((rule: string) => ignores.add(rule));
+
   const tasks = dirs.map(async folder => {
     const fileStats = await lStat(folder);
     // Check if symlink and exclude if requested
@@ -207,45 +212,24 @@ export async function collectIgnoreRules(
     const snykFiles = allIgnoredFiles.filter(f => f.endsWith(DOTSNYK_FILENAME));
     const snykExcludeRules = union(...snykFiles.map(parseFileIgnores));
 
-    // Parse rules from all .[*]ignore files inside this directory.
+    snykExcludeRules.map((rule: string) => excludes.add(rule));
+
+    // Parse rules from relevant .[*]ignore files inside this directory.
+    // Exclude ignore files under paths excluded by .snyk files.
     const ignoreFiles = allIgnoredFiles.filter(
-      // Exclude .snyk files (parsed separately above) and files in paths excluded by .snyk files.
-      f => !f.endsWith(DOTSNYK_FILENAME) && multimatch(f, snykExcludeRules).length === 0,
+      f => !f.endsWith(DOTSNYK_FILENAME) && multimatch([nodePath.dirname(f)], snykExcludeRules).length === 0,
     );
+    const ignoreFileRules = union(...ignoreFiles.map(parseFileIgnores));
 
-    return parseIgnoreRulesFromFiles(ignoreFiles, snykExcludeRules);
+    ignoreFileRules.map((rule: string) => ignores.add(rule));
   });
 
-  const customRules = await Promise.all(tasks);
+  await Promise.all(tasks);
 
-  return union(fileIgnores, ...customRules);
-}
-
-/**
- * Parse and collect ignore rules from .[*]ignore files.
- * Removes redundant rules for efficiency.
- * Removes negative rules that are superseded by .snyk exclude rules.
- */
-function parseIgnoreRulesFromFiles(ignoreFiles: string[], snykExcludeRules: string[]): string[] {
-  // Read ignore files and merge patterns.
-  const localIgnoreRules = union(...ignoreFiles.map(parseFileIgnores));
-
-  const rulesAfterProcessing: string[] = [];
-  return union(localIgnoreRules, snykExcludeRules).filter((rule: string) => {
-    // Identify and remove negative rules that are superseded by .snyk exclude rules.
-    // i.e. if a directory "foo" is ignored by a .snyk exclude rule, then a negative match
-    // for "!foo" from another ignore file is superseded and needs to be removed.
-    const negativeRuleIsSupersededBySnykExcludes =
-      rule.startsWith('!') && snykExcludeRules.some(existingRule => multimatch(rule.slice(1), existingRule).length > 0);
-    // Deduplicate rules, such as "**/dir" and "sub/dir".
-    // In this case "sub/dir" is a subset of the first rule and thus redundant.
-    const ruleIsRedundant = rulesAfterProcessing.some(existingRule => multimatch(rule, existingRule).length > 0);
-
-    const ruleIsValid = !negativeRuleIsSupersededBySnykExcludes && !ruleIsRedundant;
-    if (ruleIsValid) rulesAfterProcessing.push(rule);
-
-    return ruleIsValid;
-  });
+  return {
+    excludes: Array.from(excludes),
+    ignores: Array.from(ignores),
+  };
 }
 
 export function determineBaseDir(paths: string[]): string {
@@ -265,10 +249,14 @@ async function* searchFiles(
   patterns: string[],
   cwd: string,
   symlinksEnabled: boolean,
-  ignores: string[],
+  policies?: FilePolicies,
 ): AsyncGenerator<string | Buffer> {
-  const positiveIgnores = ignores.filter(rule => !rule.startsWith('!'));
-  const negativeIgnores = ignores.filter(rule => rule.startsWith('!')).map(rule => rule.substring(1));
+  const positiveIgnores = policies
+    ? [...policies.excludes, ...policies.ignores.filter(rule => !rule.startsWith('!'))]
+    : [];
+  const negativeIgnores = policies
+    ? policies.ignores.filter(rule => rule.startsWith('!')).map(rule => rule.substring(1))
+    : [];
   // We need to use the ignore rules directly in the stream. Otherwise we would expand all the branches of the file system
   // that should be ignored, leading to performance issues (the parser would look stuck while analyzing each ignored file).
   // However, fast-glob doesn't address the negative rules in the ignore option correctly.
@@ -292,16 +280,14 @@ async function* searchFiles(
   // `node_module/` <= ignores everything in a `node_module` folder and it's relative subfolders
   // `!node_module/my_module/` <= excludes the `my_module` subfolder from the ignore
   // `node_module/my_module/build/` <= re-includes the `build` subfolder in the ignore
-  // We feed any positive ignore rules that also match patterns back into the matcher to ensure any paths matching our
-  // patterns for supported files that were marked as ignored are not re-surfaced by the second pass.
-  const positivePatternIgnores = positiveIgnores.filter((rule: string) => isMatch(rule, deepPatterns));
   if (negativeIgnores.length) {
     const negativeSearcher = fg.stream(negativeIgnores, {
       ...fgOptions,
       cwd,
       followSymbolicLinks: symlinksEnabled,
       baseNameMatch: false,
-      ignore: positivePatternIgnores,
+      // Exclude rules should still be respected
+      ignore: policies?.excludes ?? [],
     });
     for await (const filePath of negativeSearcher) {
       if (isMatch(filePath.toString(), deepPatterns)) yield filePath;
@@ -316,7 +302,7 @@ async function* searchFiles(
 export async function* collectBundleFiles({
   symlinksEnabled = false,
   baseDir,
-  fileIgnores,
+  filePolicies,
   paths,
   supportedFiles,
 }: CollectBundleFilesOptions): AsyncGenerator<FileInfo | string> {
@@ -342,7 +328,7 @@ export async function* collectBundleFiles({
   // Scan folders
   const globPatterns = getGlobPatterns(supportedFiles);
   for (const folder of dirs) {
-    const searcher = searchFiles(globPatterns, folder, symlinksEnabled, fileIgnores);
+    const searcher = searchFiles(globPatterns, folder, symlinksEnabled, filePolicies);
     // eslint-disable-next-line no-await-in-loop
     for await (const filePath of searcher) {
       const fileInfo = await getFileInfo(filePath.toString(), baseDir, false, cache);
@@ -355,7 +341,7 @@ export async function* collectBundleFiles({
 
   // Sanitize files
   if (files.length) {
-    const searcher = searchFiles(filterSupportedFiles(files, supportedFiles), baseDir, symlinksEnabled, fileIgnores);
+    const searcher = searchFiles(filterSupportedFiles(files, supportedFiles), baseDir, symlinksEnabled, filePolicies);
     for await (const filePath of searcher) {
       const fileInfo = await getFileInfo(filePath.toString(), baseDir, false, cache);
       // dc ignore AttrAccessOnNull: false positive, there is a precondition with &&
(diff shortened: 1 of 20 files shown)


src/files.ts
```diff
@@ -187,19 +187,48 @@ export async function collectIgnoreRules(
...
-    return union(...localIgnoreFiles.map(parseFileIgnores));
+
+    // Read ignore files and merge patterns.
+    const localIgnoreRules = union(...localIgnoreFiles.map(parseFileIgnores));
+
+    // Final array of rules after processing to return.
+    const rulesAfterProcessing: string[] = [];
+
+    for (const rule of localIgnoreRules) {
+      // Identify and remove negative rules that are superseded by top-level .snyk rules.
```
strassl-snyk, line 212: q: I wonder if we could make this simpler and more resilient by inverting the logic, similar to how scm-bundle-store does it. I'm not completely familiar with how the logic here works though. I'm thinking:
- Collect the .gitignore rules
- Apply the .gitignore rules (positive and negative)
- Apply the .snyk excludes

This could be implemented by returning an object with the different kinds of rules from this function, and applying them in sequence in the `collectBundleFiles` function.

The existing design where we force everything into a single set of ignore rules seems contrary to the domain logic of having .snyk excludes take precedence, and we are hacking around it with this rule-filtering logic.

strassl-snyk (commented):
I love the test quality. One question/point of discussion as to whether we could come up with a more elegant solution.

shinin-light (approved):
I like the idea of applying an exclusion priority (i.e. `root .snyk file` > `anything else`), but I'm a bit skeptic about not having a total order of files/rules and instead relying on a single exception (i.e. the `root .snyk file`). Overall, I agree this is a minimalistic change that would bear the maximum result for the UX, which is to give back control to the user and define a clear triaging strategy when something gets weird.

src/files.ts
novalex: I was on the fence about this for a while, but eventually settled on this approach because 1) it is simple - we keep the same behaviour as before, where the "collect ignore rules" function returns a single array of rules to be used, and 2) returning different sets of rules and applying them in sequence requires either yet another pass of `searchFiles`, or significant rewriting of the file collection logic, neither of which seems worth it to me.

novalex:
Agree with your point @shinin-light , the handling of ignores and exclusions can certainly be improved further, but as it stands now we have a significant bug where our main exclusion mechanism does not do what it's supposed to do, bringing with it support tickets and annoyed users, so I'm leaning towards an iterative approach rather than a big bang fix that will potentially take much longer.

src/files.ts
```diff
@@ -187,19 +187,48 @@ export async function collectIgnoreRules(
     // Check if symlink and exclude if requested
     if (!fileStats || (fileStats.isSymbolicLink() && !symlinksEnabled) || fileStats.isFile()) return [];
 
-    // Find ignore files inside this directory
+    // Parse top-level .snyk file inside this directory
+    const topLevelSnykIgnores = parseFileIgnores(`${folder}/${DOTSNYK_FILENAME}`);
```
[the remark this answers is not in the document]
novalex: Yes, while the issues we've had reported all seemed to be related to the root `.snyk` file, any nested `.snyk` files should also be included in the initial exclude rule collection step.
At the same time, running a full recursive `.snyk` file collection might adversely impact durations, and we would still need to respect the excludes from parent directories (i.e. not collect `.snyk` files from subfolders which are marked excluded from higher-level `.snyk` files), which could complicate the logic significantly.
Will investigate, for the sake of correctness.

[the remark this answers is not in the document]
novalex: Implemented support for nested .snyk files by splitting the process into 4 main steps:
- global `.snyk` and `.[*]ignore` file collection
- `.snyk` file exclude rule parsing
- `.[*]ignore` file ignore rule parsing
- filtering and merging ignore and exclude rules

The only caveat with this approach is that it does not account for .snyk files that are excluded by other .snyk files. In order to support that we'd need to do 2 separate passes of `fast-glob`, which practically doubles the duration, or otherwise implement some mechanism for dynamically applying ignore rules while recursively collecting files, which no library seems to support.

Considering most 3rd party packages will not contain `.snyk` files, or if they do, they would only contain positive excludes and not negative ones like `.[*]ignore` files can, I don't see any risk of this adversely impacting the behaviour (e.g. unexpectedly overriding higher-level excludes, or excluding paths from higher-level directories they have no business excluding).

cc @shinin-light as it also relates to one of your points ("relying on a single exception").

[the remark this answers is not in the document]
strassl-snyk: I'll give it a proper review today or tomorrow (probably tomorrow).

>The only caveat with this approach is that it does not account for .snyk files that are excluded by other .snyk files. In order to support that we'd need to do 2 separate passes of fast-glob, which practically doubles the duration, or otherwise implement some mechanism for dynamically applying ignore rules while recursively collecting files, which no library seems to support.

q: Aren't we already doing 2 passes here? One for collecting the ignore files and one for collecting the files? Doing a 1.5th pass just over the ignore files shouldn't be expensive (but see below).

We probably don't need to address the following in this iteration, but we may want to start questioning some fundamental assumptions in this design e.g.:
- Should we even be using fast-glob at all, or should we just have your own implementation for walking the directory tree with filtering? It seems we are doing a lot of work to make the behavior fit fast-glob.
- Is the collect rules/apply rules 2 pass design the right choice or should we just be doing a single pass "collect files" step for performance reasons?

> Considering most 3rd party packages will not contain .snyk files, or if they do, they would only contain positive excludes and not negative ones like .[*]ignore files can, I don't see any risk of this adversely impacting the behaviour (e.g. unexpectedly overriding higher-level excludes, or excluding paths from higher-level directories they have no business excluding).

Do I understand correctly that you are referring to this case (dummy syntax for ease of readability):
```
# /.snyk

exclude:
 - /sub/.snyk
```

```
# /sub/.snyk

exclude:
  - ./foo.txt
```

If I understand correctly, the "expected behavior" you point out is that the exclude from `/.snyk` should invalidate the exclude from `/sub/.snyk`. In your implementation this wouldn't happen and `/sub/foo.txt` would get excluded. 

I think, if that is the only divergence, that is acceptable and  I'm not even sure whether we would even want `.snyk` files to exclude other `.snyk` files.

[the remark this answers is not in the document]
shinin-light: > In order to support that we'd need to do 2 separate passes of fast-glob, which practically doubles the duration, or otherwise implement some mechanism for dynamically applying ignore rules while recursively collecting files

There is a third option that is less fs-heavy and more computationally heavy. Instead of creating a list of "path exclusion" we could create a `map` of "exclusion source file path" to "excluded path". If we can have a correctly ordered list of all the exclusion rules (positive and negatives both added to the same list) we could definitely parse the rules in order and check their source to know if they should apply or not (based on previous rules in the list). Once the synthesis is done, we don't need any additional fast-glob pass.

> Should we even be using fast-glob at all

It's very fast, although a bit inflexible to our desires

> Is the collect rules/apply rules 2 pass design the right choice or should we just be doing a single pass "collect files" step for performance reasons?

It's a 2-passes design like it always was because the first pass is to collect all the ignore rules (`collectIgnoreRules` function calls fast-glob internally) and another pass is to collect all the valid file paths using the computed file exclusions + the extension filters coming from the API. The 1-pass design is theoretically possible, but we would need to re-write the whole file-scraping logic to either:
- add what @novalex is referring above as "dynamically applying ignore rules while recursively collecting files", which fast-glob just doesn't allow us to do (afaik)
- run a single pass by both creating the `map` mentioned above and also collect a list of `all the file paths` and then run in memory all the exclusion rule computations AND the respective application to the paths list. I'm not sure this is performant tho, since `fast-glob` is specifically designed to be very fast (and not memory intensive) when you know what you're looking for (i.e. using filters), while trying to create a full description of the customer's FS in memory might be too slow for big repos (thousands of ignored paths/dependencies). We could still make a compromise here, since most of the interesting exclusion rules are in the root, we could have 2-passes but with the first pass addressing only root files and the second pass computing the path list and exclusion rules map. This approach would be "correct" because you're not creating out-of-order exceptions to the exclusion logic since the root files are by definition the highest priority.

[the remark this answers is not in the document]
novalex: For @strassl-snyk 

> q: Aren't we already doing 2 passes here? One for collecting the ignore files and one for collecting the files? Doing a 1.5th pass just over the ignore files shouldn't be expensive (but see below).

Yes, but the first pass is only looking for ignore files, and the second one looks for all supported files but respecting the ignore rules collected in the first stage. If we were to do yet another pass to look for ignore files but only in the directories not excluded by .snyk files, it would still have to do most of the heavy-lifting to walk the entire dir structure, so in the benchmark the duration for `collectIgnoreRules` was practically doubled.

> Should we even be using fast-glob at all, or should we just have your own implementation for walking the directory tree with filtering? It seems we are doing a lot of work to make the behavior fit fast-glob.

There are a lot of libraries for this, all with drawbacks and advantages, but we're not really doing a lot of work to make it work with `fg` in particular, but rather to try to deal with the limitation of "no dynamic ignores" they all have. I'm quite sure trying to reinvent the wheel and cooking up our own solution would be more work than implementing stages and early filtering to construct the rules we provide to `fg`.

> Is the collect rules/apply rules 2 pass design the right choice or should we just be doing a single pass "collect files" step for performance reasons?

If we do a single pass without any ignore rules, the performance can actually be a lot worse, because we're looking at *all* directories in the structure for *all* supported files. Whereas with the 2-pass implementation we're only looking for supported .snyk and ignore files in all the directories, and then only looking for supported files in the directories that have not been ignored (which most times would include `node_modules` for example) so it's able to scan supported files quicker.

> Do I understand correctly that you are referring to this case (dummy syntax for ease of readability)

Yes, I was more thinking the general "exclude `/sub` from root" and in that case "stop looking at any .snyk files in there as well", but now I realise the deduplication step would remove any rules coming from `/sub/.snyk` anyway.
Your example is more salient, but as you said excluding `.snyk` files themselves is probably not something we want to support, and `.snyk` excludes cannot override higher-level exclude rules like negative matches can, so I think it's a non-issue.
Though we could theoretically parse `.snyk` files in order and skip parsing files under paths that were previously excluded by an already parsed file, it would increase complexity and likely still not work 100% correctly (due to file order as returned by fg, for example), so IMO not worth doing.

For @shinin-light 

> There is a third option that is less fs-heavy and more memory/computation intensive. Instead of creating a list of "path exclusion" we could create a map of "exclusion source file path" to "excluded path". If we can have a correctly ordered list of all the exclusion rules (positive and negatives both added to the same list) we could definitely parse the rules in order and check their source to know if they should apply or not (based on previous rules in the map). Once the synthesis is done, we don't need any additional fast-glob pass.

True, but I think guaranteeing file order is not going to be trivial, and also we'd still need to parse all ignore files first. Given a contrived example of a `.snyk` excluded directory having 100 ignore files, we'd still need to loop and parse through all of those 100 files only to collect the ignore rules that will be discarded in the filtering step. THat's why IMO looking at only `.snyk` files first and then the other ignore files not excluded by `.snyk` is about as efficient we can get, without doing more advanced wizardry like dynamic ignores as we loop through files (and that has it's own challenges including the "guaranteeing file order" one).

> run a single pass by both creating the map mentioned above and also collect a list of all the file paths. Then in memory we can perform all the exclusion rule computations AND the respective application to the paths list. I'm not sure this is performant tho, since fast-glob is specifically designed to be very fast (and not memory intensive) when you know what you're looking for (i.e. using filters), while trying to create a full description of the customer's FS in memory might be too slow for big repos (thousands of ignored paths/dependencies).

That has the potential of killing the CLI process with large-enough repos. Even the array of supported files (respecting any exclude and ignore rules) can be large enough to cause problems like OOM (see https://snyksec.atlassian.net/browse/ZEN-796). Collecting ALL files in memory would only make it worse.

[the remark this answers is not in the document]
strassl-snyk: > Yes, but the first pass is only looking for ignore files, and the second one looks for all supported files but respecting the ignore rules collected in the first stage. If we were to do yet another pass to look for ignore files but only in the directories not excluded by .snyk files, it would still have to do most of the heavy-lifting to walk the entire dir structure, so in the benchmark the duration for collectIgnoreRules was practically doubled.

What I am saying is we could do an in-memory pass just over the ignore files after we have collected them, not a full pass over the filesystem.

> There are a lot of libraries for this, all with drawbacks and advantages, but we're not really doing a lot of work to make it work with fg in particular, but rather to try to deal with the limitation of "no dynamic ignores" they all have. I'm quite sure trying to reinvent the wheel and cooking up our own solution would be more work than implementing stages and early filtering to construct the rules we provide to fg.

> If we do a single pass without any ignore rules, the performance can actually be a lot worse, because we're looking at all directories in the structure for all supported files. Whereas with the 2-pass implementation we're only looking for supported .snyk and ignore files in all the directories, and then only looking for supported files in the directories that have not been ignored (which most times would include node_modules for example) so it's able to scan supported files quicker.

Yes we are. This entire PR (and also the previous problematic behavior) is working around the fact that fast-glob doesn't fully map to our use case. When I say "single pass" I mean what you refer to as "dynamic applying excludes" - we traverse the file system once (in a single pass) and apply the excludes during the traversal. There really is no wizardry around it - it should be relatively straightforward depth-first traversal - but, as you say, it is additional work to implement and maintain.

> Though we could theoretically parse .snyk files in order and skip parsing files under paths that were previously excluded by an already parsed file, it would increase complexity and likely still not work 100% correctly (due to file order as returned by fg, for example), so IMO not worth doing.

Agreed. It's questionable of whether we want to support this edge case anyway (and/or what the correct behavior is).

src/files.ts
```diff
@@ -187,22 +193,61 @@ export async function collectIgnoreRules(
...
+  // Read ignore files and merge patterns.
+  const localIgnoreRules = union(...ignoreFiles.map(parseFileIgnores));
+
+  const rulesAfterProcessing: string[] = [];
+  return union(localIgnoreRules, snykExcludeRules).filter((rule: string) => {
+    // Identify and remove negative rules that are superseded by .snyk exclude rules.
+    // i.e. if a directory "foo" is ignored by a .snyk exclude rule, then a negative match
+    // for "!foo" from another ignore file is superseded and needs to be removed.
+    const negativeRuleIsSupersededBySnykExcludes =
+      rule.startsWith('!') && snykExcludeRules.some(existingRule => multimatch(rule.slice(1), existingRule).length > 0);
```
strassl-snyk, line 239: q: Is it really correct to match rules against each other like this? I'm thinking something like this:

Directory tree:
```
root/
  sub/
    exclude.txt
```

Superseding rule A (e.g. from `.snyk`): `root/**/exclude.txt`
Superseded rule B (e.g. from `.gitignore`): `!root/sub`


If I understand the logic here correctly, for rule B we would have:
- `negativeRuleIsSupersededBySnykExcludes = false`
- `ruleIsRedundant = false`
- `ruleIsValid = true`

And therefore rule B would again incorrectly override the exclusion from rule A.

I have pushed an example test for what I mean [here](https://github.com/snyk/code-client/commit/740169a3f7ae5459a09be668de1919934f5b59e5).

tests/fixtures/file-ignores/combined/.gitignore
```diff
@@ -0,0 +1,4 @@
+/gitignore-root-excluded
```
strassl-snyk, line 1: q: Did you forget to commit the files here because the `.gitignore` excluded them? 😅

src/files.ts
```diff
@@ -187,22 +193,61 @@ export async function collectIgnoreRules(
...
+    // Deduplicate rules, such as "**/dir" and "sub/dir".
+    // In this case "sub/dir" is a subset of the first rule and thus redundant.
+    const ruleIsRedundant = rulesAfterProcessing.some(existingRule => multimatch(rule, existingRule).length > 0);
+
+    const ruleIsValid = !negativeRuleIsSupersededBySnykExcludes && !ruleIsRedundant;
+    if (ruleIsValid) rulesAfterProcessing.push(rule);
```
strassl-snyk, line 245: enh: Maintaining an array of "rules we have processed" inside a filter means this entire process depends a lot on iteration order (and is a bit of an anti-pattern). A better approach would be to make this a two step process:
- Filter out the redundant rules
- Filter out the redundant rules based on the previous results

q: Is the redundancy check even necessary?

strassl-snyk (changes requested):

src/files.ts
novalex: Good catch, will look into this case, however your commit is adding a new test (marked broken, but which passes), and removes the snapshot contents from the previous test which fails. Apparently `matchInlineSnapshot` does not like the existing snapshots with dynamic path prefixes and just removes them.

tests/fixtures/file-ignores/combined/.gitignore
novalex: Nope, we're testing the rule collection in isolation so don't actually care about any project files in there, only the exclude and ignore files.

src/files.ts
strassl-snyk: I was just pushing it as an example to make it clearer. The test passes, but the output is incorrect, right?

tests/fixtures/file-ignores/combined/.gitignore
strassl-snyk: Ahh gotcha, makes sense

src/files.ts
novalex: Ah, I see. So the rule collection seems correct to me, as we shouldn't drop the rule for `!root/sub` entirely, but because of our 2-pass implementation with negative patterns second we'll still see that `/root/sub/exclude.txt` in the bundle files.

One thing that comes to mind is feeding the positive ignore rules back into the `negativeSearcher` fg call, which should ensure that ignored files are really ignored and not surfaced by the second pass, but unsure what the implications are then.

Will think about this some more.
cc @shinin-light , if you have any thoughts 🙏

shinin-light: > feeding the positive ignore rules back into the negativeSearcher fg call, which should ensure that ignored files are really ignored and not surfaced by the second pass

I'm pretty sure I've tried something like that already and it failed. I remember playing around both filters and positive/negative ignores both in the 1-pass and 2-pass scenarios since `fg` allows you to have both patterns and exclusions, but in the end I remember stopping after successfully applying "negated filters" as "excludes" in the second pass. Feel free to experiment more

novalex: Hmmm, yeah, we'd probably have to limit the ignores for the second pass to only the .snyk rules, not all of them, otherwise negative .gitignore matchers would still not work...

novalex: I think I found a solution, that also works for the `TODO` case we had with the `/not/ignored/this_should_be_ignored.jsx` file:
On the second pass with the negative patterns we provide the positive ignore rules that also match one of our supported files patterns.
This should ensure files under paths which have been ignored and then overridden by a negative rule are surfaced, but specific files we support that have been marked ignored are still ignored in the second pass.
It's not perfect, but IMO it improves correctness a lot and should minimise edge cases we have.

src/files.ts
novalex: The redundant rule filtering is not really critical and meant to reduce the number of ignore patterns we give the library, i.e. remove more specific rules if a general rule is available that would match anyway. It functionally doesn't matter if we process these our of order and might miss removing redundant rules, the more important part is filtering out those superseded by `.snyk` rules which are collected beforehand.

src/files.ts
novalex: And a breaking case:
- A `.snyk` file excluding `/gitignore-unexcluded/sub`, and
- A `.gitignore` file with a negative rule for `!/gitignore-unexcluded`
Will still surface files from `/gitignore-unexcluded/sub`, so not correct.

After discussing with @strassl-snyk it looks like our best option (short of writing our own lib implementation) is to only feed in `.snyk` rules to the second pass `ignore`.
Some diagrams that may help understanding these flows [here](https://miro.com/app/board/uXjVMnCevuQ=/?share_link_id=301693422540).
Will be working on the "Other proposal" changes now which should address all edge cases so far as well as the breaking case described above.

src/files.ts
novalex: This filtering was removed in the latest implementation.

src/files.ts
novalex: Updated implementation:
- Rule filtering was removed, we now parse all `.snyk` files, and then parse all other ignore files from paths that were not excluded.
- File ignores are now returned as a `FilePolicies` object with separate `excludes` and `ignores`, so in the file collection step we can apply all positive ignores (combined), and only the `.snyk` excludes in the second pass for negative patterns.

The case described above is fixed, but the previous case with the `/not/ignored/this_should_be_ignored.jsx` file is back 😢 
So we still don't correctly apply negative patterns, but we do ensure that `.snyk` excludes always take precedence and that negative patterns will never override excludes.

src/files.ts
```diff
@@ -220,10 +249,14 @@ async function* searchFiles(
   patterns: string[],
   cwd: string,
   symlinksEnabled: boolean,
-  ignores: string[],
+  policies?: FilePolicies,
```
strassl-snyk, line 252: q: Why is this optional? Shouldn't it always be passed?

strassl-snyk (commented):
Generally LGTM. One required question and a few enhancements.

PS: Your PR description is out of data

src/interfaces/analysis-options.interface.ts
```diff
@@ -46,10 +46,15 @@ export interface AnalyzeFoldersOptions {
...
+export interface FilePolicies {
+  excludes: string[];
+  ignores: string[];
+}
+
 export interface CollectBundleFilesOptions extends AnalyzeFoldersOptions {
   supportedFiles: SupportedFiles;
   baseDir: string;
-  fileIgnores: string[];
+  filePolicies?: FilePolicies;
```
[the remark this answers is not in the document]
novalex: The interface and `collectBundleFiles` are both internal and not used outside of `code-client` as far as I can see. The main external methods are `createBundleWithCustomFiles` and `createBundleFromFolders`, and the interfaces for those have not changed.

strassl-snyk (approved):

src/files.ts
```diff
@@ -177,30 +177,59 @@ function generateAllCaseGlobPattern(fileExtension: string): string {
...
+
+    // Parse rules from all .snyk files inside this directory.
+    const snykFiles = allIgnoredFiles.filter(f => f.endsWith(DOTSNYK_FILENAME));
+    const snykExcludeRules = union(...snykFiles.map(parseFileIgnores));
+
+    snykExcludeRules.map((rule: string) => excludes.add(rule));
+
+    // Parse rules from relevant .[*]ignore files inside this directory.
+    // Exclude ignore files under paths excluded by .snyk files.
+    const ignoreFiles = allIgnoredFiles.filter(
```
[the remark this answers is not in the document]
novalex: If we're excluding the directories these files are under via `.snyk` there's no reason to parse and collect rules for them, they'd be relative to the excluded path and irrelevant/redundant.

src/files.ts
novalex: Wasn't sure whether it should be or not, it's not really needed for this to work, but might be cleaner to have it required than doing null checks/coalescing. Will change.

src/files.ts
```diff
@@ -177,30 +177,59 @@ function generateAllCaseGlobPattern(fileExtension: string): string {
...
   symlinksEnabled = false,
   fileIgnores: string[] = IGNORES_DEFAULT,
-): Promise<string[]> {
+): Promise<FilePolicies> {
+  const excludes: Set<string> = new Set();
+  const ignores: Set<string> = new Set();
+
+  fileIgnores.map((rule: string) => ignores.add(rule));
+
   const tasks = dirs.map(async folder => {
```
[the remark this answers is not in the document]
novalex: I wouldn't say this is a hard pattern to understand, it's just initialising empty arrays and populating them in a loop. Getting the task results and then doing merges to get the final object was not pretty either.
Kind of on the fence but considering.

src/files.ts
```diff
@@ -177,30 +177,59 @@ function generateAllCaseGlobPattern(fileExtension: string): string {
...
         cwd: folder,
         followSymbolicLinks: symlinksEnabled,
       },
     );
-    // Read ignore files and merge new patterns
-    return union(...localIgnoreFiles.map(parseFileIgnores));
+
+    // Parse rules from all .snyk files inside this directory.
+    const snykFiles = allIgnoredFiles.filter(f => f.endsWith(DOTSNYK_FILENAME));
+    const snykExcludeRules = union(...snykFiles.map(parseFileIgnores));
```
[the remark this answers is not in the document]
novalex: `union` has the effect of removing duplicates as well, `flatMap` does not. Leaving as-is, unless you see a good reason to get rid of `union`?

src/files.ts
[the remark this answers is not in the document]
novalex: Refactored collection to avoid mutating inside the tasks.

src/files.ts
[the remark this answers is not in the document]
novalex: Haha actually nevermind, we still have [tests](https://app.circleci.com/pipelines/github/snyk/code-client/837/workflows/9ad3b679-67dd-4088-9b99-0e78bda1cd1e/jobs/1937) running on Node 10 and `flatMap` is not supported there 😅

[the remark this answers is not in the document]
strassl-snyk: Well that settles it than. Fair point on the duplicates too, I didn't think about that being intentional behavior.

PR #202, opened 2023-09-11
feat: add routing to GW option for sast

Noa-Savransky:
- [X] Ready for review
- [X] Linked to Jira ticket - https://snyksec.atlassian.net/browse/PUL2-801

#### What does this PR do?
Adding logic to route to gateway instead of the deeproxy service

src/utils/httpUtils.ts

You can run the tests

For a temporary stage where some Deeproxy changes only deployed to FedRAMP we add this logic to use the new Deeproxy arch' in the FedRAMP clusters. 
When the changes are deployed to all the clusters we can use this new logic as default logic.

### DIFF
--- a/src/utils/httpUtils.ts
+++ b/src/utils/httpUtils.ts
@@ -1,0 +1,13 @@
+export function getURL(baseURL: string, path: string, orgId?: string): string {
+  if (routeToGateway(baseURL)) {
+    if (!orgId) {
+      throw new Error('Org is required for this operation');
+    }
+    return `${baseURL}/hidden/orgs/${orgId}/code${path}`;
+  }
+  return `${baseURL}${path}`;
+}
+
+function routeToGateway(baseURL: string): boolean {
+  return baseURL.includes('snykgov.io');
+}
--- a/tests/httpUtils.spec.ts
+++ b/tests/httpUtils.spec.ts
@@ -1,0 +1,33 @@
+import { getURL } from "../src/utils/httpUtils";
+
+describe('getURL', () => {
+    it('should return base + path if not fedramp', () => {
+        const base = 'api.dev.snyk.io';
+        const path = '/analysis';
+
+        const result = getURL(base, path);
+
+        expect(result).toEqual(base+path);
+    });
+
+    it('should return base + org routing + path if fedramp', () => {
+        const base = 'api.snykgov.io';
+        const path = '/analysis';
+        const orgId = '1-2-3-4'
+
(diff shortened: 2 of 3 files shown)


src/utils/httpUtils.ts
```diff
@@ -0,0 +1,13 @@
...
+    if (!orgId) {
+      throw new Error('Org is required for this operation');
+    }
+    return `${baseURL}/hidden/orgs/${orgId}/code${path}`;
+  }
+  return `${baseURL}${path}`;
+}
+
+function routeToGateway(baseURL: string): boolean {
+  return baseURL.includes('snykgov.io');
```
vasser, line 12: So, when all deeproxy changes will be rolled out to all envs, this check will become redundant?

src/utils/httpUtils.ts
```diff
@@ -0,0 +1,13 @@
+export function getURL(baseURL: string, path: string, orgId?: string): string {
+  if (routeToGateway(baseURL)) {
+    if (!orgId) {
+      throw new Error('Org is required for this operation');
+    }
+    return `${baseURL}/hidden/orgs/${orgId}/code${path}`;
+  }
+  return `${baseURL}${path}`;
```
vasser, line 8: nit. I'd consider  using `path.join` here to avoid juggling with slashes, etc. Then it would not make any difference if you pass base url that ends with / and path that starts with /, eg:

```
path.join('https://snykgov.io/', '/test') => 'https://snykgov.io/test'
```
and 
```
path.join('https://snykgov.io', 'test') => 'https://snykgov.io/test'
```

tests/httpUtils.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  it('should throw an error if fedramp and org is missing', () => {
+    const base = 'api.snykgov.io';
+    const path = '/analysis';
+
+    try {
+      const result = getURL(base, path);
+    } catch (err) {
+      expect(err).toBeDefined();
+    }
```
vasser, line 31: suggestion: let's make this test more specific, since there can be any other failure and we are not checking that the error thrown is the on we are expecting

```suggestion
    expect(() => getURL(base, path)).toThrowError('Org is required for this operation');
```

vasser (approved):

src/utils/httpUtils.ts
Noa-Savransky: Yes.
We will remove this check and use the new way for all the routes.
Old versions will still call with the old way but we have a plan on how we support the 2 situations

PR #204, opened 2023-09-29
fix: add orgId for filters endpoint as well

jozsef-armin-hamos:
- [ ] Ready for review
- [ ] Linked to Jira ticket

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -34,6 +34,7 @@ export interface ConnectionOptions {
   source: string;
   requestId?: string;
   org?: string;
+  orgId?: string;
   extraHeaders?: { [key: string]: string };
 }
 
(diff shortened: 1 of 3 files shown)


src/http.ts
```diff
@@ -180,12 +180,13 @@ export async function getFilters(
   source: string,
   attempts = MAX_RETRY_ATTEMPTS,
   requestId?: string,
+  orgId?: string,
```
florinMiro, line 183: What if org id is missing?

jozsef-armin-hamos: If we need a route to the gateway, because we are on fedramp, it will throw an error. If not fedramp then nothing, it won't use org for the URL generation.

florinMiro (approved):

iainjmitchell (approved):
As OrgId is optional, this should work fine :) 
@florinMiro double checked too.

iainjmitchell:
asked about testing, but is only manually tested and code-client is 🤞 not likely to survive long-term

PR #209, opened 2023-10-20
fix: add extra check to validate org id

jozsef-armin-hamos:
- [ ] Ready for review
- [PUL2-898] Linked to Jira ticket

#### What does this PR do?

Adds a regex validation to the orgId, so invalid id's will fail in the library, instead of failing in the api-gateway.


src/utils/httpUtils.ts
```diff
@@ -11,3 +13,7 @@ export function getURL(baseURL: string, path: string, orgId?: string): string {
 function routeToGateway(baseURL: string): boolean {
   return baseURL.includes('snykgov.io');
 }
+
+function isValidOrg(orgId?: string): boolean {
```
strassl-snyk, line 17: enh: `orgId: string | undefined` - there isn't much of a reason of supporting `isValidOrg()`

src/utils/httpUtils.ts
```diff
@@ -11,3 +13,7 @@ export function getURL(baseURL: string, path: string, orgId?: string): string {
...
+  return Boolean(orgId && ORG_ID_REGEXP.exec(orgId));
```
strassl-snyk, line 18: q: Why the boolean constructor here?

strassl-snyk, line 18: enh: `ORG_ID_REGEXP.test` may be more along the lines of what you are looking for?

src/utils/httpUtils.ts
```diff
@@ -1,6 +1,8 @@
+import { ORG_ID_REGEXP } from '../constants';
+
 export function getURL(baseURL: string, path: string, orgId?: string): string {
   if (routeToGateway(baseURL)) {
-    if (!orgId) {
+    if (!isValidOrg(orgId)) {
       throw new Error('Org is required for this operation');
```
strassl-snyk, line 6: enh: The error message here may be a bit misleading - the user did provide an org id but the error implies it is missing.

src/utils/httpUtils.ts
```diff
@@ -1,6 +1,8 @@
+import { ORG_ID_REGEXP } from '../constants';
+
 export function getURL(baseURL: string, path: string, orgId?: string): string {
   if (routeToGateway(baseURL)) {
-    if (!orgId) {
```
strassl-snyk, line 3: q: I dimly recall there being some support for passing org slugs instead of IDs. Is this related/what are the implications?

strassl-snyk (commented):
What's the reasoning behind this change? I can't quite determine the context from the information in the ticket and the PR?

src/utils/httpUtils.ts
jozsef-armin-hamos: We are validating the existence of the orgId, in a FedRamp environment. We also want to validate the orgId format, in case someone uses code client, and sends the org name instead, to not send a request all the way to the api-gateway, because it will fail anyway.

src/utils/httpUtils.ts
jozsef-armin-hamos: Fair enough, I corrected it to `A valid Org id is required`

src/utils/httpUtils.ts
strassl-snyk: Hmm so what are the implications there? I assume the API gateway doesn't support org slugs anymore? I understand this is only FedRAMP right now, but does that mean that if we were to, in theory, enable it for non-FedRAMP environments we would no longer support org slugs?

src/utils/httpUtils.ts
strassl-snyk: Was this meant as a reply to this? :)

>What's the reasoning behind this change? I can't quite determine the context from the information in the ticket and the PR?

What I'm trying to understand is why the client-side validation here is necessary - is it just for better error messaging?

src/utils/httpUtils.ts
jozsef-armin-hamos: Exactly. api-gateway routing only supports org id's, and not the slugs. Once all the other environments will be using the api-gateway as a sole means of entry to the cluster, we will require this check everywhere, unless api-gateway changes the way it handles incoming requests (enables org slugs).

src/utils/httpUtils.ts
jozsef-armin-hamos: Indeed, a better error handling is the main reason. If an incorrect org id reaches the api-gateway, the error is not very clear, and hard to debug the reason of failure.

src/utils/httpUtils.ts
strassl-snyk: Okay, but to be clear: We can't do that in existing environments. If my understanding is correct, we would have a bunch of customers relying on org slugs working and the API gateway migration would break all of them. We'd have to make sure we transform the org slug into an org id before we make the request.

Not a blocker for this PR specifically, but it may be worth document some of the discussions we had here in a comment :)

strassl-snyk (approved):
No blockers left, only improvements

src/utils/httpUtils.ts
jozsef-armin-hamos: In the existing environments the routing doesn't use orgId at all. We are normally sending requests to endpoints such as `/filters` or `/bundle`, directly to deeproxy. The api-gateway expects the request url to contain the orgId always. This is the reason the url is created differently for FedRamp. Once the api-gateway will be used for other environments, the url will have to be constructed in the same way.

jozsef-armin-hamos: Also, as far as I know, org slugs are not used right now for request url generation, but I may be wrong. For sure they are not used for code analysis in the code-client library.

strassl-snyk: I am aware, I was talking about once we migrate the existing environments :)

src/utils/httpUtils.ts
jozsef-armin-hamos: Do you prefer casting the value to boolean with `!!`?

strassl-snyk: Why would we need to cast it at all? Shouldn't we just be able to do (roughly):

```
return orgId !== undefined && ORG_ID_REGEXP.test(orgId)
```

jozsef-armin-hamos: You are right, since orgId can be undefined or a string. I just thought it is safer to ensure that orgId has a value, and is not null before doing the regex check. But that should not be a problem when using typescript.

strassl-snyk: You can also do the following, if you want to catch the case. But if that is a possible situation, you should probably update the type definitions too :)

```
return orgId != null && ORG_ID_REGEXP.test(orgId)
```

PR #210, opened 2023-11-07
feat: add error handling for limit errors

nick-watson:
- [x] Ready for review
- [ ] Linked to Jira ticket

#### What does this PR do?

- Adds error handling for catalogue (json api format) errors to throw errors with a detail
- Handles analysis and create/extend bundle requests to detail [limit errors](https://docs.snyk.io/more-info/error-catalog#code) thrown by SAA and FBS
- Will be used in [this](https://github.com/snyk/cli/pull/4922) PR for the CLI to display error to the user
- User will see error message + user friendly detail + link to error catalogue

#### Screenshots
Screenshot 2023-11-08 at 17 17 05

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -3,26 +3,28 @@ import pick from 'lodash.pick';
 import { gzip } from 'zlib';
 import { promisify } from 'util';
 
-import { ErrorCodes, GenericErrorTypes, DEFAULT_ERROR_MESSAGES, MAX_RETRY_ATTEMPTS } from './constants';
+import { DEFAULT_ERROR_MESSAGES, ErrorCodes, GenericErrorTypes, MAX_RETRY_ATTEMPTS } from './constants';
 
 import { BundleFiles, SupportedFiles } from './interfaces/files.interface';
 import { AnalysisResult, ReportResult } from './interfaces/analysis-result.interface';
 import { FailedResponse, makeRequest, Payload } from './needle';
 import {
-  AnalysisOptions,
   AnalysisContext,
+  AnalysisOptions,
   ReportOptions,
   ScmReportOptions,
 } from './interfaces/analysis-options.interface';
-import { getURL } from './utils/httpUtils';
+import { generateJsonApiError, getURL, isJsonApiErrors } from './utils/httpUtils';
 
 type ResultSuccess<T> = { type: 'success'; value: T };
-type ResultError<E> = {
+
+export type ResultError<E> = {
   type: 'error';
   error: {
-    statusCode: E;
+    statusCode: E | number;
     statusText: string;
     apiName: string;
+    detail?: string | undefined;
   };
 };
 
@@ -41,24 +43,26 @@ export interface ConnectionOptions {
 // The trick to typecast union type alias
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function isSubsetErrorCode<T>(code: any, messages: { [c: number]: string }): code is T {
-  if (code in messages) {
-    return true;
-  }
-  return false;
+  return code in messages;
 }
 
 function generateError<E>(
   errorCode: number,
   messages: { [c: number]: string },
   apiName: string,
-  error?: string,
+  errorMessage?: string,
+  errors?: unknown,
 ): ResultError<E> {
+  if (isJsonApiErrors(errors)) {
+    return generateJsonApiError<E>(errors, errorCode, apiName);
+  }
+
   if (!isSubsetErrorCode<E>(errorCode, messages)) {
-    throw { errorCode, messages, apiName };
+    throw { statusCode: errorCode, statusText: errorMessage || 'unknown error occurred', apiName };
   }
 
   const statusCode = errorCode;
-  const statusText = error ?? messages[errorCode];
+  const statusText = errorMessage ?? messages[errorCode];
 
   return {
     type: 'error',
@@ -92,8 +96,7 @@ interface StartSessionOptions {
 export async function compressAndEncode(payload: unknown): Promise<Buffer> {
   // encode payload and compress;
   const deflate = promisify(gzip);
-  const compressedPayload = await deflate(Buffer.from(JSON.stringify(payload)).toString('base64'));
-  return compressedPayload;
+  return await deflate(Buffer.from(JSON.stringify(payload)).toString('base64'));
 }
 
 export function startSession(options: StartSessionOptions): StartSessionResponseDto {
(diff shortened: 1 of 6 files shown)


src/http.ts
```diff
@@ -3,26 +3,28 @@ import pick from 'lodash.pick';
...
 import { FailedResponse, makeRequest, Payload } from './needle';
 import {
-  AnalysisOptions,
   AnalysisContext,
+  AnalysisOptions,
   ReportOptions,
   ScmReportOptions,
 } from './interfaces/analysis-options.interface';
-import { getURL } from './utils/httpUtils';
+import { generateJsonApiError, getURL, isJsonApiErrors } from './utils/httpUtils';
```
metju90, line 17: Does `snykApiErrors` sound better than `jsonApiErrors` ?

src/http.ts
```diff
@@ -3,26 +3,28 @@ import pick from 'lodash.pick';
...
-type ResultError<E> = {
+
+export type ResultError<E> = {
   type: 'error';
   error: {
-    statusCode: E;
+    statusCode: E | number;
     statusText: string;
     apiName: string;
+    detail?: string | undefined;
```
novalex, line 27: `q`: Can we avoid the new `detail` field? Unless we want to show them separately somehow, `statusText` can also include the "friendly" error message currently, so we can do the same with the `jsonApiError` detail message - i.e. `error.title` + `error.detail` (+ link) under `statusText`.

src/http.ts
```diff
@@ -41,24 +43,26 @@ export interface ConnectionOptions {
...
+  errorMessage?: string,
+  errors?: unknown,
 ): ResultError<E> {
+  if (isJsonApiErrors(errors)) {
+    return generateJsonApiError<E>(errors, errorCode, apiName);
+  }
+
   if (!isSubsetErrorCode<E>(errorCode, messages)) {
-    throw { errorCode, messages, apiName };
+    throw { statusCode: errorCode, statusText: errorMessage || 'unknown error occurred', apiName };
```
novalex, line 61: `q`: Not sure how this is handled, we're throwing this but it's not clear how downstream callers deal with it, are we just going to the global error handler in the CLI for example?

nick-watson: I don't think the previous implementation ever worked, when it is thrown the CLI just output `[Object object]` in the console as the thrown object was used as the message in the [default error handling](https://github.com/snyk/cli/blob/f88d296daaf3b09d236f34af5008542522454f66/src/lib/plugins/sast/index.ts#L123). This make's the error handle-able by the [code client error type](https://github.com/snyk/cli/blob/f88d296daaf3b09d236f34af5008542522454f66/src/lib/plugins/sast/index.ts#L136) in the CLI.

src/needle.ts
```diff
@@ -45,10 +45,12 @@ interface SuccessResponse<T> {
   success: true;
   body: T;
 }
+
 export type FailedResponse = {
   success: false;
   errorCode: number;
   error: Error | undefined;
+  errors?: unknown;
```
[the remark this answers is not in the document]
nick-watson: I was thinking more defensively as `makeRequest` is a generic util and the response could change in the future (with older clients in mind). We could move the validation up from the `generateError` function?

src/http.ts
nick-watson: I'm not sure as the detail is how we differentiate which errors should be shown to the user Vs hardcoded user messaging [here](https://github.com/snyk/cli/blob/f88d296daaf3b09d236f34af5008542522454f66/src/lib/plugins/sast/index.ts#L167). The statusText is only used in the [debugging output](https://github.com/snyk/cli/blob/f88d296daaf3b09d236f34af5008542522454f66/src/lib/plugins/sast/index.ts#L125) AFAICT.

src/needle.ts
```diff
@@ -130,7 +135,7 @@ export async function makeRequest<T = void>(
       await sleep(REQUEST_RETRY_DELAY);
     } else {
       attempts = 0;
-      return { success: false, errorCode, error };
+      return { success: false, errorCode, error, errors };
```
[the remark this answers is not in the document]
nick-watson: If we move the json api validation into `makeRequest` we can just rename to `jsonApiErrors`?

src/http.ts
```diff
@@ -3,26 +3,28 @@ import pick from 'lodash.pick';
...
+import { generateJsonApiError, getURL, isJsonApiErrors } from './utils/httpUtils';
 
 type ResultSuccess<T> = { type: 'success'; value: T };
-type ResultError<E> = {
+
+export type ResultError<E> = {
   type: 'error';
   error: {
-    statusCode: E;
+    statusCode: E | number;
```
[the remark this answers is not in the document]
nick-watson: Admittedly this was path of least resistance as the generic type is passed down through functions. Will take another look.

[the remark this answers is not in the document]
nick-watson: updated

src/needle.ts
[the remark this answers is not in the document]
nick-watson: updated to be more specific

src/needle.ts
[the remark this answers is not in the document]
nick-watson: updated to be more specific

src/http.ts
novalex: Ah crap we have that custom handling in the CLI too, right. 🙏

src/needle.ts
```diff
@@ -109,6 +115,13 @@ export async function makeRequest<T = void>(
 
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
     const errorMessage = response?.body?.error;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-member-access
+    const errors = response?.body?.errors;
+
+    if (isJsonApiErrors(errors)) {
+      jsonApiError = errors[0];
```
novalex, line 122: `nit`: This behaviour (only handling first error) is not documented anywhere.
I don't see cases where we have more than one error from the API TBH, but know the JSON API requires arrays.
If someone were to change the API to actually return more than one error they will probably not realise the client can only handle one though.

nick-watson: Yeah I guess this is a limitation right now, where would you think we can document it?

src/http.ts
nick-watson: `jsonApiErrors` is descriptive of the type we are checking for I think but updated the function name.

src/needle.ts
metju90: The good thing is if we have more than 1 error, nothing breaks. They will be ignored.

Perhaps a single comment from the service that throws the error?

metju90 (approved):

novalex (approved):

PR #216, opened 2023-11-17
fix: add posibility to add extra headers to filters endpoint

jozsef-armin-hamos:
- [ ] Ready for review
- [PUL2-771] Linked to Jira ticket

#### What does this PR do?

Getting the filters works without Authorization header in all environments. In FedRamp, this is not allowed, so I am addig the possibility to use the extra headers when getting the file extension in code-client.
This is required for the e2e test orb to work with FedRamp-Alpha.

[PUL2-771]: https://snyksec.atlassian.net/browse/PUL2-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/bundles.ts
+++ b/src/bundles.ts
@@ -220,9 +220,17 @@ export async function getSupportedFiles(
   requestId?: string,
   languages?: string[],
   orgId?: string,
+  extraHeaders?: Record<string, string>,
 ): Promise<SupportedFiles> {
   emitter.supportedFilesLoaded(null);
-  const resp = await getFilters(baseURL, source, MAX_RETRY_ATTEMPTS, requestId, orgId);
+  const resp = await getFilters({
+    baseURL,
+    source,
+    orgId,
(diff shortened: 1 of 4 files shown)


strassl-snyk (approved):
Does this mean we moving towards sending auth headers for /filters requests in all instances (FedRAMP and non-FedRAMP)? I'm all for that (there is no reason for filters to not require authn). However, this will probably require changes in multiple clients.

jozsef-armin-hamos:
CLI is overwriting the headers anyway, so practically this does not affect it (tested it locally). AFAIK the only repo that directly uses code-client is the e2e-test-orb. For now I am only adding optional extraHeaders, but indeed in the future we will probably require all endpoints to use Authorization headers.

